PR #3, opened 2018-02-24
Added two custom fields

xenonis:
Added two custom fields to be copied. Pls refer to wpml ticket https://wpml.org/forums/topic/translated-post-does-not-copy-page-settings-follow-up/

### DIFF
--- a/divi-builder/wpml-config.xml
+++ b/divi-builder/wpml-config.xml
@@ -15,6 +15,8 @@
         <custom-field action="copy">_et_pb_predefined_layout</custom-field>
         <custom-field action="copy">_et_pb_page_layout</custom-field>
         <custom-field action="copy">_et_pb_side_nav</custom-field>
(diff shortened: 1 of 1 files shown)


dgwatkins:
this is a duplicate of https://github.com/OnTheGoSystems/wpml-config/pull/2
closing here

xenonis:
@dgwatkins thanks for accepting and merging #3. 
I added this change for [Divi (theme)](https://github.com/OnTheGoSystems/wpml-config/blob/master/divi/wpml-config.xml) and for the [Divi Builder](https://github.com/OnTheGoSystems/wpml-config/blob/master/divi-builder/wpml-config.xml), hence the duplicate (I think).

Are they copies of each other, so it's only needed for one of them?

thanks
JP

dgwatkins:
i didnt notice that, thanks for the notice

dgwatkins:
Its merged now, thanks

PR #7, opened 2018-02-27
Revert "[Enfold] Update xlm config file"

amElnagdy:
@InoPlugs
we're setting `_aviaLayoutBuilderCleanData`  to ignore and creating it programmatically.

To achieve what you asking, so you can use the 'Duplicate' option. Or you can use WPML -> Translation Management -> Custom XML Configuration and override our online config file.

### DIFF
--- a/enfold/wpml-config.xml
+++ b/enfold/wpml-config.xml
@@ -7,9 +7,9 @@
         <taxonomy translate="1">portfolio_entries</taxonomy>
     </taxonomies>
     <custom-fields>
-        <custom-field action="copy-once">_aviaLayoutBuilder_active</custom-field>
-        <custom-field action="copy-once">_avia_builder_shortcode_tree</custom-field>
-        <custom-field action="copy-once">_aviaLayoutBuilderCleanData</custom-field>
+        <custom-field action="copy">_aviaLayoutBuilder_active</custom-field>
+        <custom-field action="copy">_avia_builder_shortcode_tree</custom-field>
+        <custom-field action="ignore">_aviaLayoutBuilderCleanData</custom-field>
     </custom-fields>
     <language-switcher-settings>
         <key name="icl_lang_sel_config">


vatoyiit:
@amElnagdy
Not doing that breaks our layout when users want to have different shortcode structure per language and asking users to override config file is not a viable option for us. 
I'm afraid we're going to need to find another solution.

amElnagdy:
@vatoyiit having different shortcode structures per languages can be done (actually it should be done) using the manual translation method as described here https://wpml.org/documentation/translating-your-contents/page-builders/translating-content-created-with-page-builders-using-manual-translation/

No need to use WPML Translation Editor.

InoPlugs:
@amElnagdy  @vatoyiit 

Sorry, but this is not correct.

Using only the base plugin (no other WPML plugin activated): 

It can be reproduced even using the duplicate as described above. When I tested it with my changes (copy once) it worked. We update the shortcode tree _avia_builder_shortcode_tree on every page update from _aviaLayoutBuilderCleanData or post_content (on non ALB pages).

The reason to copy once is on creating the page to have the start content.

- Enable ALB debug mode (you can use the show parser info button below the debug window (since 4.2.1) any time to check stored shortcode tree)
- Create a new page
- Add ALB elements
- publish
- check "duplicate" in sidebar and create translation
- click pencil to edit translation
- Add an ALB element to the translation
- Update
- Check Shortcode Parser Button - still the same tree as base page (should be the tree from updated translation, but was overwritten again)
- Modify the ALB structure in base page
- Update
- Check Shortcode Parser Button in base page - modified structure
- Check Shortcode Parser Button in translated page - same tree as in base page (not the structure of the translated page)

amElnagdy:
@InoPlugs this is not true :)

You have to click '**Translate Independently**' before updating the translated page.

As you can see, I'm able to add elements to the translated page without problems (only WPML, no additional addons)

en
fr


Also, I have added an icon element to the main language page and added a header to the translated page after that and all working fine.

additional-en

additional-fr

If you can reproduce your issue on an installation and want me to investigate it, you can use Duplicator and pass to me the installer and zip files.

InoPlugs:
@amElnagdy 

I checked again and clicked the button "Translate independently" before updateing the translated page.

The problem is not the meta field _aviaLayoutBuilderCleanData. That is correct.

The problem is _avia_builder_shortcode_tree. This field must reflect the shortcode structure of the page.

And this meta field is updated always to the structure of the main page.

You can check the stored array either in the database or when clicking this button and tab shortcode tree:

parser

amElnagdy:
@InoPlugs Let me ask,

Setting `_avia_builder_shortcode_tree` to copy-once, should fix the issue?

We have no problems setting it to 'copy-once'

InoPlugs:
@amElnagdy 

I think it should. 

When the page is updated we hook into the save_post action  with prio 10 and update this meta field from post_content (non ALB pages) or _aviaLayoutBuilderCleanData.

Your "copy" setting seems to overwrite our changes again.

amElnagdy:
Perfect!
As I mentioned, we have no problems with setting this field `_avia_builder_shortcode_tree` to **copy-once**.

On it 👍

InoPlugs:
@amElnagdy 

Thanks. We will test it, when your done.

amElnagdy:
@InoPlugs you can try this before we make it public for everyone.

Activate WPML Translation Management and go to WPML -> Translation Management -> Custom XML Configuration and add the following

 ```
<wpml-config>
 <custom-fields>
       <custom-field action="copy-once">_avia_builder_shortcode_tree</custom-field>
 </custom-fields>
</wpml-config>
```

If it's fixing the issue, we'll implement it immediately.

**Comment by InoPlugs** [comment]
@amElnagdy 

Will do and we will come back when done.

Thx.

**Comment by InoPlugs** [comment]
@amElnagdy 

```
<wpml-config>
<custom-fields>
      <custom-field action="copy-once">_avia_builder_shortcode_tree</custom-field>
</custom-fields>
</wpml-config>
```

works fine if only base plugin is activated and when editing pages directly (when the save_post filter is fired).

When using translation managment and editing/saving a translation the save_post filter is not used by your plugin. But we can hook into "wpml_pro_translation_completed" action to update the shortcode tree.

As long as your plugins have no other updates to the post content that do not fire the save_post filter I think we can add the this to your config file (but we should do this when we release our next Enfold update).

There is only one "limitation":

All translations must be either ALB or non ALB pages - mixing is not possible because you copy _aviaLayoutBuilder_active.

Is this a "limitation" or can we think (in future) about allowing to mix this?

amElnagdy:
Thanks, @InoPlugs for your tests.

As for `_aviaLayoutBuilder_active`, I don't think setting it to 'copy-once' will be a problem. It'll need to be tested of course.

If you want this to be tested in the future, it's OK. When you release Enfold's upcoming release, please let me know and I'll set `_avia_builder_shortcode_tree` to 'copy-once'

Thanks!

InoPlugs:
@amElnagdy 

Hi, Kriesi released Enfold  v4.2.6 right  now to Theme Forest. So it should be online soon.

Can you now update _avia_builder_shortcode_tree to 'copy-once' please.

Thanks and have a nice day.

amElnagdy:
@InoPlugs just created it https://github.com/OnTheGoSystems/wpml-config/pull/10 :)

PR #20, opened 2018-05-11
Added The7 WPBakery Page Builder

srdjan-jcc:
Dream Theme forked plugin The7 WPBakery Page Builder should use same config as WPBPB. Config for custom cells can be found in theme's config file.


amElnagdy:
Regarding the id, should not it be the plugin's folder name?
@dgwatkins @srdjan-jcc

dgwatkins:
Yes, @srdjan-jcc does this plugin have a different folder name?

Good catch Ahmed

amElnagdy:
@dgwatkins it's `the7-js_composer` as far as I can see.

PR #34, opened 2018-06-28
updating avada and wcml

amElnagdy:
for wcml -> https://onthegosystems.myjetbrains.com/youtrack/issue/wcml-2461


amElnagdy:
@dgwatkins :)

PR #35, opened 2018-07-02
Add admin-texts for Exit Popup

dgwatkins:
https://wordpress.org/plugins/exit-popup/


exit-popup/wpml-config.xml
```diff
@@ -0,0 +1,8 @@
+<wpml-config>
+    <admin-texts>
+        <key name="exit_popup_popup_title_color"/>
```
srdjan-jcc: color?

dgwatkins: Removed, thanks!

PR #38, opened 2018-07-17
comp-3043 - Cleanup remote WPML config files

amElnagdy:


avada/wpml-config.xml
```diff
@@ -204,35 +204,15 @@
         <shortcode>
             <tag>fusion_builder_column</tag>
             <attributes>
-                <attribute>background_image</attribute>
-                <attribute type="link">link</attribute>
```
dgwatkins: Are you sure this line should be removed?
I think it should stay, please confirm @amElnagdy

avada/wpml-config.xml
```diff
@@ -342,11 +314,9 @@
         <shortcode>
             <tag>fusion_lightbox</tag>
             <attributes>
-                <attribute>title</attribute>
-                <attribute type="link">href</attribute>
```
dgwatkins: same here

avada/wpml-config.xml
amElnagdy: @dgwatkins it's duplicated. The shortcode is duplicated.

avada/wpml-config.xml
amElnagdy: These attributes don't exist. I added the correct ones.

PR #39, opened 2018-07-18
Remove non-text elements to avoid translators making mistakes

dgwatkins:
- Unify whitespaces (add ?w=1 to the URL to ignore whitespace changes).
- Remove `vc_raw_html` and `contact-form-7`
- Add missing title attribute to some vc_wp_xxxxx shortcodes
- Remove taxonomy from tag cloud shortcode
- Remove url/items from RSS shortcode
- Synchronize these changes to modified versions of VC

### DIFF
--- a/js_composer/wpml-config.xml
+++ b/js_composer/wpml-config.xml
@@ -1,7 +1,7 @@
 <wpml-config>
-	<custom-fields>
-		<custom-field action="ignore">_vc_post_settings</custom-field>
-	</custom-fields>
+    <custom-fields>
+        <custom-field action="ignore">_vc_post_settings</custom-field>
+    </custom-fields>
     <shortcodes>
         <shortcode>
             <tag>vc_column_text</tag>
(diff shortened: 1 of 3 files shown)


js_composer/wpml-config.xml
```diff
@@ -334,11 +347,5 @@
                 <attribute>share_text_custom_text</attribute>
             </attributes>
         </shortcode>
-	<shortcode>
-            <tag>contact-form-7</tag>
-            <attributes>
-                <attribute>id</attribute>
```
amElnagdy: @dgwatkins how can the user add the translated form ID then?

amElnagdy: the same for the URL by the way

dgwatkins: - Bottom line is that the translator won't know the id to translate it. We will have to add some code somewhere.
- Let me put the url back,.

amElnagdy: CF7 is different because it's not translated.

dgwatkins:
@amElnagdy 
After todays meeting, we agreed:
- to remove the url until 4.1.0 because its causing more problems than solving
- keep the contact form id while we work on a better solution.
Agreed?

PR #52, opened 2018-09-18
[for 4.1.0] Add config for media shortcodes in Divi

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5819

### DIFF
--- a/divi-builder/wpml-config.xml
+++ b/divi-builder/wpml-config.xml
@@ -1,6 +1,6 @@
 <wpml-config>
     <built-with-page-builder>/\[(et_pb_section|et_pb_row|et_pb_column)/</built-with-page-builder>
-	<shortcode-list>et_pb_section,et_pb_row,et_pb_column,et_pb_accordion_item,et_pb_cta,et_pb_map_pin,et_pb_text,et_pb_blurb,et_pb_circle_counter,et_pb_contact_field,et_pb_contact_form,et_pb_image,et_pb_number_counter,et_pb_post_nav,et_pb_pricing_table,et_pb_tab,et_pb_toggle,et_pb_counter,et_pb_button,et_pb_countdown_timer,et_pb_login,et_pb_team_member,et_pb_post_slider,et_pb_search,et_pb_slide,et_pb_signup,et_pb_testimonial,et_pb_fullwidth_post_slider,et_pb_fullwidth_header,et_pb_fullwidth_image,et_pb_fullwidth_portfolio,et_pb_audio,et_pb_line_break_holder,et_pb_blog,et_pb_fullwidth_post_title,et_pb_code</shortcode-list>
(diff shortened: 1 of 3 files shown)


divi-builder/wpml-config.xml
```diff
@@ -248,5 +260,23 @@
                 <attribute>album_name</attribute>
             </attributes>
         </shortcode>
+        <shortcode>
+            <tag media-only="1">et_pb_*</tag>
```
brucepearson: @strategio I'm a bit wary about the `*` here. I think it's going to be confusing that this will only work for media and not for strings unless you also plan to make it work for strings.

divi-builder/wpml-config.xml
```diff
@@ -248,5 +260,23 @@
...
+            <media-attributes>
+                <media-attribute type="url">background_image</media-attribute>
+            </media-attributes>
+        </shortcode>
+        <shortcode>
+            <tag media-only="1">et_pb_video_slider_item|et_pb_video</tag>
```
brucepearson: @strategio Same comment as above but with the `|` format. I don't really see an issue of defining two separate tags in the wpml-config.xml instead of combining like you have done here.

divi-builder/wpml-config.xml
strategio: @brucepearson, in Divi all the block can have a background image. That's why I decided to use this pattern.

Trying to maintain a definition for each shortcode is not viable IMO.

I don't if this is transposable to strings (and how complex it is) but I can create a ticket if you feel a need.

divi-builder/wpml-config.xml
strategio: OK @brucepearson, we can define 2 shortcodes in that case.

divi-builder/wpml-config.xml
brucepearson: @strategio I would prefer to have a definition for every shortcode. I know it's more work to create the definition file but it's less work that has to happen inside WPML code.

strategio: OK @brucepearson, I added this `background_image` attribute to all existing shortcodes. We might miss some of the shortcodes, but it's probably not the most important image to convert.

amElnagdy:
I think this will need to wait too @strategio? 
Just for the record, changes made to Divi and Divi Builder should be done to Extra theme as well.

strategio:
I re-wrote the media settings in the 3 config files (this it ready for review now).

divi-builder/wpml-config.xml
```diff
@@ -49,7 +49,7 @@
         <shortcode>
             <tag>et_pb_accordion_item</tag>
             <attributes>
-                <attribute>title</attribute>
+                <attribute>title</attribute><attribute type="media-url">background_image</attribute>
```
brucepearson: @strategio Please put on a separate line.

strategio: Corrected!

PR #58, opened 2018-10-08
wpml-config file for Forminator plugin

kkarpieszuk:
@comp-code-review please review and add


kkarpieszuk:
@dgwatkins @srdjan-jcc or anyone, please review and merge

dgwatkins:
Hi @kkarpieszuk 
You need to add the corresponding entry to config-index.xml
Thanks!

kkarpieszuk:
@dgwatkins file updated, please check. I set override flag to false because we are going to switch this preference anyway if author will be cooperating. This is just a temporary file

dgwatkins:
I have never tried setting `override_local` to false.
If we don't want an override, we usually just omit the whole attribute.
However, I dont think its important in this case because its temporary.

PR #68, opened 2019-01-10
Add CF fusion_builder_status as copy once in and element_category taxonomy as translatable in order to be able to translate a Container using default editor

adrianosferreira:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6106


adrianosferreira:
@amElnagdy @dgwatkins please review it.

PR #77, opened 2019-03-13
wpmltm-3322 Cornerstone classic custom headlines registered for translation twice

kagg-design:
`ignore-content="1"` added to Classic Custom Headline and Classic Features Headline.

Removed repetitive names from `<shortcode-list>`

Some formatting has done.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3322


kagg-design:
@OnTheGoSystems/wpml Please review

kagg-design:
@strategio yes, that is the reason. So, I have blocked registering of strings in 2 shortcodes.

strategio:
Got it @kagg-design! I think it's worth a comment in the config file otherwise it will be hard to remember why we did this in 6 months.

Could you add:

```
<!-- This block is already registered through the API, we don't want to register it twice -->
```

kagg-design:
@strategio sure, done!

PR #79, opened 2019-03-26
adding fusion dropcap

amElnagdy:


dgwatkins:
@amElnagdy its sync'd too

PR #82, opened 2019-04-08
Removed GeneratePress theme

srdjan-jcc:
Removing theme, author wants config in premium plugin GP Premium.
It's fine, since all preferences are for premium version.


srdjan-jcc:
it already includes file, but it will be up-to-date (it misses important prefs) when author finishes adding preferences and publishes new plugin version. Maybe we should wait until then and remove only after it's published?

dgwatkins:
yes, ok
thanks

PR #83, opened 2019-04-11
Beta support for Muffin page builder (betheme)

dgwatkins:


amElnagdy:
@dgwatkins if we're going to host the config file, should not we merge the config from the local file and set the online one to override the local one? Or ask the author to include this in the local file. 

Bottom line, clients should have a complete config file. What do you think?

dgwatkins:
@amElnagdy I set it to override_local

If the author decides to maintain it we can remove it from here.
Sounds good?

dgwatkins:
@amElnagdy oh, I gotcha now.
The theme already has a wpml-config.xml file.
Merging it next .... thanks

dgwatkins:
@amElnagdy Can you take a second look?
Thanks!

PR #88, opened 2019-05-21
Setting Layout Taxonomies to be not translatable

amElnagdy:


dgwatkins:
Can you sort out the tabs @amElnagdy ?
Sorry for being picky :-)

And another question, can we sync this to S3 already?

amElnagdy:
> Can you sort out the tabs @amElnagdy ?
> Sorry for being picky :-)
> 
> And another question, can we sync this to S3 already?

sorry for that, it should be OK now?
And yes, it can be synced.

PR #94, opened 2019-10-03
Handle translating link target for gutenberg image and button blocks

andreas-pa:


andreas-pa:
please @dgwatkins One review! 

Thanks!

PR #95, opened 2019-10-11
Translating the section CSS ID

amElnagdy:
This is needed for anchor links.

https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-4306


amElnagdy:
@dgwatkins nope, it's the CSS #id selector. The client can name a section and translate it

dgwatkins:
> @dgwatkins nope, it's the CSS #id selector. The client can name a section and translate it

Ok, as long as a remote translator knows how to translate it we are ok.

PR #96, opened 2019-11-04
Set _element_data to 'copy-once' for duplicates to work.

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6738

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -4,7 +4,7 @@
         <custom-field action="copy">_elementor_edit_mode</custom-field>
         <custom-field action="copy">_elementor_template_type</custom-field>
         <custom-field action="copy">_elementor_version</custom-field>
-        <custom-field action="ignore">_elementor_data</custom-field>
+        <custom-field action="copy-once">_elementor_data</custom-field>
         <custom-field action="ignore">_elementor_css</custom-field>
     </custom-fields>
     <custom-types>


elementor/wpml-config.xml
```diff
@@ -4,7 +4,7 @@
         <custom-field action="copy">_elementor_edit_mode</custom-field>
         <custom-field action="copy">_elementor_template_type</custom-field>
         <custom-field action="copy">_elementor_version</custom-field>
-        <custom-field action="ignore">_elementor_data</custom-field>
+        <custom-field action="copy">_elementor_data</custom-field>
```
amElnagdy: In the case of translation, the translated content will be overwritten with the original content if this custom field is set to copy.

dgwatkins: I did test it and I was under the impression that the "copy" happens before the "translation".

Let me double-check though ...
by the way, did you test it  @amElnagdy?

amElnagdy: Not with the recent versions. But in the past, yes I tested.

dgwatkins: I see, there is a gap between the "copy" (which happens after updating the original) and when the user completes the "translation". It's not a viable solution then.

dgwatkins: Then lets try "translate".
It seems to do the trick

amElnagdy: 'translate' will make it horrible while translating using WPML translation editors. As the content of the custom field will appear in the TE

dgwatkins: I see you have already been here :-)

dgwatkins: @amElnagdy how about removing the line completely?

amElnagdy: This 'may' be the solution. Because with normal custom fields and when you don't set a translation preference for them, WPML takes care of them while duplicating and ignore them while translating.
But this needs some real testing.

dgwatkins: I already saw a problem .. there is no way to remove the field on an existing setup.

dgwatkins: I know this sounds like a joke by now, but seriously ... "copy-once" is going to be the solution.

dgwatkins: @amElnagdy I tried both workflows (creating and updating) with ATE, CTE and duplication and this works perfectly.

Can you please confirm this and if it works for you I will ask the WPML QA team to take a look as well.

Thanks!

amElnagdy: Sure, I'll give it a go. What I'm going to check:

- An already translated page, change the custom field and try to update it.

- New page translation. (CTE, ATE, and manual)

amElnagdy: @dgwatkins it seems to be working correctly. Tested CTE, ATE, manual and duplication.

dgwatkins: @amElnagdy I take it duplication worked too?

Can you approve this?

andreasciamanna:
@OnTheGoSystems/wpml

dgwatkins:
@andreasciamanna Its a long story, the problem is actually that when we set it to ignore it's not copied when duplicating the post. This is the only way it actually works.

andreasciamanna:
>  Its a long story, the problem is actually that when we set it to ignore it's not copied when duplicating the post. This is the only way it actually works.

I get that. I just want to be sure that, by doing that, we won't cause other kinds of issues.

dgwatkins:
@andreasciamanna Yes, I am worried about the same too.
But I did extensive testing and so did Ahmed.

I am currently running codeception tests to make sure it all continues to run fine.

... and I want to run this ticket through QA too, even if it's just a wpml-config.xml change.

Any other ideas to avoid other kinds of issues?

dgwatkins:
@brucepearson The copy-once happens first, and after the translated values are updated.

PR #98, opened 2019-11-07
Copy post meta to flag the PB editor in Beaver Builder, Cornerstone and Elementor

strategio:
This is required because if the PB editor is disabled, we won't run all
the logic when the translation job is saved (and so the post meta wouldn't be
copied).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6929


strategio:
@OnTheGoSystems/wpml

PR #103, opened 2019-12-02
Copy post meta to flag the PB editor in Beaver Builder, Cornerstone & Elementor

dgwatkins:
This is required because if the PB editor is disabled, we won't run all
the logic when the translation job is saved (and so the post meta wouldn't be
copied).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6929

### DIFF
--- a/config-index.xml
+++ b/config-index.xml
@@ -2,6 +2,8 @@
 	<plugins>
 		<item id="10bit-woocommerce-gateway-yaadpay" override_local="true">10bit WooCommerce Gateway Yaadpay</item>
 		<item id="book-now">Book Now</item>
+		<item id="beaver-builder" override_local="true">Beaver Builder Plugin (Lite Version)</item>
+		<item id="beaver-builder" override_local="true">Beaver Builder Plugin (Pro Version)</item>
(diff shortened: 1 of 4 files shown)


config-index.xml
```diff
@@ -2,6 +2,8 @@
 	<plugins>
 		<item id="10bit-woocommerce-gateway-yaadpay" override_local="true">10bit WooCommerce Gateway Yaadpay</item>
 		<item id="book-now">Book Now</item>
+		<item id="beaver-builder" override_local="true">Beaver Builder Plugin (Lite Version)</item>
+		<item id="beaver-builder" override_local="true">Beaver Builder Plugin (Pro Version)</item>
```
andreasciamanna: @dgwatkins can we have two items with the same ID? Can't we use a single item with just `Beaver Builder Plugin` as a value?

dgwatkins: yes, no problem.

We do that when required

andreasciamanna: My main concern is having two identical IDs: that may cause some validation issues, but you know better than me if that's the case.

PR #104, opened 2019-12-03
Handle Divi theme builder header/body/footer layouts

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6888


gordonsh24:
@dgwatkins +1

PR #114, opened 2020-03-12
Cannot translate dynamic fields in Avada 6.2.

kagg-design:
A number of elements in Avada can have dynamic content. I have examined all of them and added `dynamic_params` attribute to relevant shortcodes.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7156


kagg-design:
@comp-code-review Please review.

PR #115, opened 2020-03-12
Adjust our wpml-config.xml to the latest wpml-config.xml supplied with Avada theme.

kagg-design:
https://onthegosystems.myjetbrains.com/youtrack/issue/comp-3433


kagg-design:
@comp-code-review

PR #120, opened 2020-04-01
WIP Improve `built-with-page-builder` regex to also match global modules

strategio:
The current regular expressions are too restrictive for some page
builders, it matches only the layout shortcode that usually wrap the
content, but some global module do not have any wrapper.

@amElnagdy, @dgwatkins, I looked at this [commit (comp-3075- Built with Page Builder)](https://github.com/OnTheGoSystems/wpml-config/commit/4f517acd93cc34cee33f5f0090aeb047cd528bbb) to retrieve the files we need to update. But this commit is quite old now and we probably added compatibility to other page builders. Could you tell which one we also need to update?

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7193


amElnagdy:
@strategio nothing else needs updating. 

I'm wondering, what about the files hosted with other themes? For example, UX Builder (Flatsome)

strategio:
@andreasciamanna sure, I'll improve the commit message.

> @strategio nothing else needs updating.
> 
> I'm wondering, what about the files hosted with other themes? For example, UX Builder (Flatsome)

@amElnagdy, can't we have a remote config file for these? How many page builder plugins/themes do we have? (I am talking about shortcodes page builders).

Note that the problem would only affect global elements, and not all page builders are affected. So if you could provide a list of shortcode page builders that are compatible with WPML, I could check this.

One more question, I could see in our config files that some are using the `shortcode` property but do not have `built-with-page-builder` and `shortcode-list` (I found "pro", "x" and "brooklyn"). We should also update it, right?

dgwatkins:
1) can't we have a remote config file for these?

We can add as many remote files as we want, but this means we have to maintain them from now on. 

The config-index file is also sent twice a day to every WPML install out there, so we try to keep it as lean as possible to keep the bills down and the speed up.

So if its a small theme, or the author is collaborative, or the author requests it, we leave it to him.

2) I dont think we support global elements for X/Pro (which both use cornerstone) or brooklyn

Do we still need to add it?

strategio:
After discussion with David, I will change the approach to build the regular expression with declared shortcodes in the config file.

amElnagdy:
@strategio well as it needed for the global elements, I think we're covered here.

strategio:
Closing here!

I will revert part of my changed that created this defect (temporarily disabled) and I will think of another solution.

PR #129, opened 2020-06-08
Copying shortcodes custom CSS to the translation

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlsupp-7884

### DIFF
--- a/js_composer/wpml-config.xml
+++ b/js_composer/wpml-config.xml
@@ -6,6 +6,7 @@
 	</custom-types>
     <custom-fields>
         <custom-field action="ignore">_vc_post_settings</custom-field>
+        <custom-field action="copy">_wpb_shortcodes_custom_css</custom-field>
     </custom-fields>
     <shortcodes>
         <shortcode>


cburkitt:
Hello OTGS

Why have you made this change?!

The `_wpb_shortcodes_custom_css` is part of the **page content**. It contains timestamps. Consider this shortcode from one of my pages:

`[vc_row css=".vc_custom_1497444459383{margin-top: 40px !important;}"]`

which WP Bakery parses into custom css which outputs in the `<head>` of the page:

`<style>.vc_custom_1497444459383{margin-top: 40px !important;}</style>`.

I have many translated pages which are very different from the originals. I do not copy the content directly from the original to the translated version. I edit the rows, the text boxes, many things. Now, thanks to your change, all the shortcode css in my translated pages is **ignored** because WPML is copying the postmeta from the original! So my translated pages are using incorrect css.

Did you understand the consequences of this change?

Please explain how I am to correct this problem. I do not use (and have not installed) the Translation Management module — I have no need for it.

Thank you,

Charles

amElnagdy:
Hi Charles,

We made this change because while using WPML Advanced Translation Editor, the custom CSS made by the builder was not copied over to the translated pages.

Having different styles for translations is not common. 

If you're using Translation Management module, it'd be easy to override this setting for this custom field. But as far as I can see, you don't want to use WPML Translation Management plugin so I'll set this custom field to `copy-once`, it'll be copied to the translated page and you'll be able to change its settings and values.

I'll open a new Merge Request for it.

cburkitt:
Thank you Ahmed.

Alternatively, perhaps the settings for Custom Fields could be available in
WPML > Settings all the time instead of only when the Translation
Management plugin is installed.

I installed the Translation Management plugin on my development site and
found a new link on the main settings page (Show system fields) which
revealed the _wpb_shortcode_css setting. I am not sure why this setting is
hidden here, only for users of this optional plugin.

I am afraid that I found the Translation Management plugin very troublesome
and I prefer not to have it installed.

Thank you very much for your help,

Charles

PR #132, opened 2020-06-13
WPBakery PB - Allow changing shortcodes custom CSS

amElnagdy:
See https://github.com/OnTheGoSystems/wpml-config/pull/129 comments.

### DIFF
--- a/js_composer/wpml-config.xml
+++ b/js_composer/wpml-config.xml
@@ -6,7 +6,7 @@
 	</custom-types>
     <custom-fields>
         <custom-field action="ignore">_vc_post_settings</custom-field>
-        <custom-field action="copy">_wpb_shortcodes_custom_css</custom-field>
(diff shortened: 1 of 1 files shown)


dgwatkins:
We can't merge this because we would break updates when using ATE or CTE.

Instead, we need to copy this field only when ATE or CTE are used.

I have created a ticket for WPML core so we can resolve this properly.

Meanwhile, affected clients (using manual translation) can create a wpml-config.xml file with these contents and place it in the js_composer folder:
```
<wpml-config>
<custom-fields>
<custom-field action="ignore">_wpb_shortcodes_custom_css</custom-field>
</custom-fields>
</wpml-config>
```

andreasciamanna:
Since it's can't be merged, can we close this PR please?

dgwatkins:
Closing because it will be handled with code in VC's compatibility class (wpmlcore-7364)

PR #134, opened 2020-06-18
Copy primary category to translations.

dgwatkins:
The term_id is auto-adjusted by WPML.


dgwatkins:
@OnTheGoSystems/compatibility

PR #136, opened 2020-06-25
Removing the encoding for Avia team member link

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-5025


amElnagdy:
the link is not encoded

PR #142, opened 2020-07-15
Select sub-keys from _fusion custom field for translation

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-5093


amElnagdy:
@dgwatkins @strategio I left the old custom fields as they were for the sake of backward compatibility.

PR #144, opened 2020-07-20
Add "Click to Chat" plugin config

vukvukovich:


amElnagdy:
@dgwatkins this needs to be synced?

dgwatkins:
Lets discuss this on Wednesdays meeting

amElnagdy:
@dgwatkins I'm not sure what to discuss. What I meant is does anything need to be synced after the merge?

dgwatkins:
@amElnagdy It does need to be sync'd but I am not sure if we should add every little plugin just for a pipeline to pass earlier. We need to set some requirements for adding items here or it will soon get out of hand (ie: popularity threshold?).

amElnagdy:
@dgwatkins I understand. We are already doing this. This is the [plugin page](https://wordpress.org/plugins/click-to-chat-for-whatsapp/) on WordPress.org

PR #148, opened 2020-07-31
UAG config update

vukvukovich:


kkarpieszuk:
@vukvukovich could you change indetations to tabs? will be easier to see what you have changed

to do this quickly with phpstorm: Edit > Convert indents > To tabs

ping me on MM when done, I don't get notifications from github :/

PR #153, opened 2020-09-25
Added translation for the blog posts block

bfintal:
Added translation for the read more button


vukvukovich:
@bfintal I see that we already have readmore key included. Is this not working?

vukvukovich:
Closed since no reply from the author.

PR #157, opened 2020-10-27
Removing the config files that are being handled by the authors.

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/ma-4168


izzi-hassan (approved):
Looks good to me!

dgwatkins (approved):
LGTM

PR #159, opened 2020-10-30
Update config for Custom Post Type UI

strategio:
Made the option `cptui_taxonomies` translatable in the same way as `cptui_post_type` + Added the ability to translate the other labels of these custom types.

https://onthegosystems.myjetbrains.com/youtrack/issue/comp-3502


dgwatkins (approved):
:+1: @strategio

PR #162, opened 2020-12-15
Update config file for Sensei LMS

donnapep:
This PR updates the config file for Sensei LMS to be the same as https://github.com/Automattic/sensei/pull/3840. In testing, our local version seemed to be overriding the hosted version, but thought it would be safest to make the same updates here.

This PR:
- Enables translating modules, questions and question categories.
- Enables translating user-defined values specified in settings.

It's not clear to me why `quiz`, `multiple_question` and `sensei_message` were previously set to translatable. Messages are sent between teachers and students, so those don't need to be translated, while `quiz` and `multiple_question` aren't directly accessible in WP admin.

### DIFF
--- a/config-index.xml
+++ b/config-index.xml
@@ -20,7 +20,7 @@
 		<item id="nw-adcart-for-woocommerce" override_local="true">NW ADCart for WooCommerce</item>
 		<item id="paypal-for-woocommerce" override_local="true">PayPal for WooCommerce</item>
 		<item id="product-enquiry-pro-for-woocommerce" override_local="true">Product Enquiry Pro for WooCommerce</item>
-		<item id="sensei" override_local="true">Sensei</item>
+		<item id="sensei">Sensei LMS</item>
 		<item id="subscriptio">Subscriptio</item>
 		<item id="sitepress-multilingual-cms" override_local="true">WPML Multilingual CMS</item>
 		<item id="the-events-calendar" override_local="true">The Events Calendar</item>
--- a/config-index.xml
+++ b/config-index.xml
@@ -20,6 +20,7 @@
 		<item id="nw-adcart-for-woocommerce" override_local="true">NW ADCart for WooCommerce</item>
 		<item id="paypal-for-woocommerce" override_local="true">PayPal for WooCommerce</item>
 		<item id="product-enquiry-pro-for-woocommerce" override_local="true">Product Enquiry Pro for WooCommerce</item>
+		<item id="sensei">Sensei</item>
 		<item id="sensei">Sensei LMS</item>
 		<item id="subscriptio">Subscriptio</item>
 		<item id="sitepress-multilingual-cms" override_local="true">WPML Multilingual CMS</item>
--- a/sensei/wpml-config.xml
+++ b/sensei/wpml-config.xml
@@ -10,10 +10,9 @@
 	<custom-types>
 		<custom-type translate="1">course</custom-type>
 		<custom-type translate="1">lesson</custom-type>
-		<custom-type translate="0">multiple_question</custom-type>
+		<custom-type translate="1">multiple_question</custom-type>
 		<custom-type translate="1">question</custom-type>
-		<custom-type translate="0">quiz</custom-type>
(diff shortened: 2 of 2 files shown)


kkarpieszuk:
Hi @donnapep setting `translate=0` means user would never be able to translate them. I don't remember specific case but we had scenarios when user had to do this. 

I suggest one of those approaches:
- remove the entry completely. The default translation preference is 0 but if the entry is absent in wpml-config file, user will be able to change this in wp-admin > wpml > settings > custom post types
- leave it as is but add ` display-as-translated="1"` parameter. With this, entry will be set as translatable but if translation is not provided, the original will be in use

what do you think?

kkarpieszuk:
p.s. to @strategio as far as I remember we were going to remove config files from our repository and make them stored only inside other plugins/themes directories, right? Here is Pull request to sync Sensei version of the file with file stored in our repo. Maybe all we have to do is just remove the file?

But @donnapep still, please review what I suggested

donnapep:
@kkarpieszuk Thanks. I opted to remove the one for `sensei_messages` since that should never be applicable. I also found [an old PR](https://github.com/OnTheGoSystems/wpml-config/pull/6), which seems to suggest that not being able to translate `quiz` and `multiple_question` could cause problems? I've reinstated those to 1, but am still not 100% confident in this.

kkarpieszuk:
yes, but like I suggested it could be like:

```
<custom-type translate="1" display-as-translated="1">multiple_question</custom-type>
```

This means user can translate multiple_question but if he did not provide translation, it would not be a problem: WPMl will fall back to default language's value

donnapep:
Gotcha. Updated in https://github.com/OnTheGoSystems/wpml-config/pull/162/commits/0c381646d61af21692e129535ad22a9974b53318.

kkarpieszuk (approved):

kkarpieszuk:
It is ok for me, but in ethe company we have a rule to get two reviews, I asked @strategio

strategio:
Hi @donnapep, thanks for your contribution!

Indeed, we are progressively trying to remove entries on our remote repository. In the recent years we started abusing of this feature and it has lost its initial purpose to be a fallback source when the `wpml-config.xml` file is missing in the component.

In this file https://github.com/OnTheGoSystems/wpml-config/blob/master/config-index.xml#L23, you can see that the local config file is overriden by the remote one (from here).

My suggestion is that we remove the `override_local="true"` attribute on our repo, so the local file will have priority and Sensei plugin will definitely take the full control of the WPML configuration.

Does it make sense to you @donnapep?

donnapep:
@strategio Sure thing. Please see my latest commits.

config-index.xml
```diff
@@ -20,7 +20,7 @@
 		<item id="nw-adcart-for-woocommerce" override_local="true">NW ADCart for WooCommerce</item>
 		<item id="paypal-for-woocommerce" override_local="true">PayPal for WooCommerce</item>
 		<item id="product-enquiry-pro-for-woocommerce" override_local="true">Product Enquiry Pro for WooCommerce</item>
-		<item id="sensei" override_local="true">Sensei</item>
+		<item id="sensei">Sensei LMS</item>
```
strategio, line 23: @donnapep, the value of this entry should match the [name of the plugin in its header](https://github.com/OnTheGoSystems/wpml-config#index-file-format).

If the plugin name has been changed, we need 2 entries, like this:

```xml
<item id="sensei">Sensei</item>
<item id="sensei">Sensei LMS</item>
```

That way, we'll match plugins with both names.

strategio (changes requested):

config-index.xml
donnapep: Ah, that's why my local version was overriding WPML's. 💡 https://github.com/OnTheGoSystems/wpml-config/pull/162/commits/eed8dd7d6e27de0aed64126a0bc13fe85d0a58ac

strategio (approved):
It looks good @donnapep, thanks a lot :)

@dgwatkins, could you please check and deploy this?

dgwatkins (approved):

PR #164, opened 2021-01-11
Advanced Gutenberg plugin name update

vukvukovich:


config-index.xml
```diff
@@ -72,7 +72,7 @@
 		<item id="woocommerce-mix-and-match-products" override_local="true">WooCommerce Mix and Match</item>
 		<item id="kadence-blocks" override_local="true">Kadence Blocks - Gutenberg Page Builder Toolkit</item>
 		<item id="ultimate-addons-for-gutenberg" override_local="true">Ultimate Addons for Gutenberg</item>
-		<item id="advanced-gutenberg" override_local="true">Advanced Gutenberg</item>
+		<item id="advanced-gutenberg" override_local="true">PublishPress Blocks</item>
```
dgwatkins, line 75: @vukvukovich Is this a name change?

If so, I would suggest two keep both lines during the transition period.

dgwatkins: I will go ahead and approve my own suggestion because you need this to be propagated tomorrow.

Ping me if you have any concerns.

dgwatkins (approved):
I made a small change before approving

vukvukovich:
@dgwatkins You are right, thanks!

PR #169, opened 2021-01-15
Elementor: translate the add-to-cart button label

dgwatkins:
https://github.com/OnTheGoSystems/wpml-config/compare/wpmlcore-7714

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -50,5 +50,13 @@
                 <field type="Shortcode">shortcode</field>
             </fields>
         </widget>
+        <widget name="wc-add-to-cart">
+            <conditions>
+                <condition key="widgetType">wc-add-to-cart</condition>
+            </conditions>
+            <fields>
+                <field type="Button Text" editor_type="LINE">text</field>
+            </fields>
+        </widget>
     </elementor-widgets>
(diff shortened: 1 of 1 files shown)


dgwatkins:
@OnTheGoSystems/compatibility quick review please

kkarpieszuk (approved):

elementor/wpml-config.xml
```diff
@@ -50,5 +50,13 @@
                 <field type="Shortcode">shortcode</field>
             </fields>
         </widget>
+        <widget name="wc-add-to-cart">
+            <conditions>
```
strategio, line 54: @dgwatkins, the `conditions` are optional. Here, since the widget name is the same as the condition, we don't need it.

So this should be just enough:

```xml
        <widget name="wc-add-to-cart">
            <fields>
                <field type="Button Text" editor_type="LINE">text</field>
            </fields>
        </widget>
```

strategio (approved):
It looks good @dgwatkins, I just shared a simplification.

elementor/wpml-config.xml
dgwatkins: Thanks @strategio 

I applied your suggestion

strategio (approved):

PR #172, opened 2021-01-21
Add new admin strings for Astra 3.0

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/comp-3538

### DIFF
--- a/astra/wpml-config.xml
+++ b/astra/wpml-config.xml
@@ -13,6 +13,15 @@
     <key name="astra-settings">
       <key name="footer-sml-section-1-credit"/>
       <key name="footer-sml-section-2-credit"/>
+      <key name="header-html-1"/>
+      <key name="header-html-2"/>
+      <key name="header-button1-text"/>
+      <key name="header-button1-link-option">
+        <key name="url"/>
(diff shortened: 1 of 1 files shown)


astra/wpml-config.xml
```diff
@@ -13,6 +13,14 @@
     <key name="astra-settings">
       <key name="footer-sml-section-1-credit"/>
       <key name="footer-sml-section-2-credit"/>
+      <key name="header-html-1"/>
+      <key name="header-html-2"/>
+      <key name="header-button1-text"/>
+      <key name="header-button1-link-option">
+        <key name="url"/></key>
```
dgwatkins, line 20: @strategio This ending `</key>` here doesn't look right

Is `header-html-1` really inside `astra-settings` and `footer-html-1` outside?

It doesn't validate anyway

strategio: Thanks for catching this @dgwatkins! I took my settings from the WPML Custom XML interface and it does not format in the same way... This should be fixed now!

dgwatkins (approved):
Looks good @strategio

PR #174, opened 2021-01-21
Validate XML file when running composer

dgwatkins:
@strategio Do you think we can use this in a github workflow?
Please review if you agree


strategio:
Very nice @dgwatkins, good initiative :)

Yes, we can definitely use it in github workflow.

You can merge this and I can setup the workflow if you wish.

+1

strategio (approved):

PR #176, opened 2021-01-22
Add xpath for core/table headers in sitepress-multilingual-cms 

BobbyKarabinakis:
add //th xpath for core/table so table headers are translatable in the CTE and ATE

Related WPML ticket: https://wpml.org/forums/topic/i-am-not-able-to-find-that-string-for-translation/


dgwatkins (approved):
:+1:  @BobbyKarabinakis

PR #178, opened 2021-02-03
Add support for Recent Posts widget in Elementor

amElnagdy:
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-5494

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -55,5 +55,13 @@
                 <field type="Button Text" editor_type="LINE">text</field>
             </fields>
         </widget>
+        <widget name="wp-widget-recent-posts">
+            <conditions>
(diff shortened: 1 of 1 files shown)


elementor/wpml-config.xml
```diff
@@ -55,5 +55,13 @@
                 <field type="Button Text" editor_type="LINE">text</field>
             </fields>
         </widget>
+        <widget name="wp-widget-recent-posts">
+            <conditions>
+                <condition key="widgetType">wp-widget-recent-posts</condition>
```
strategio, line 60: @amElnagdy, when the `condition` is the same as the widget `name`, you can ignore it (it will be the default condition).

amElnagdy: That's true! Fixed.

strategio (approved):
+1 @amElnagdy

PR #180, opened 2021-02-14
compsupp-5516 - changes the editor style for _et_pb_old_content

ofwald:
## Coming from compsupp-5516
The _et_pb_old_content field shows in the translation editor as a line, but it needs to be a text area/wysiwyg.

## What does the fix does?
It adds the style attribute to the field.


dgwatkins (approved):
:+1: @ofwald

PR #186, opened 2021-03-12
[Cornerstone] transpose the shortcode config to the widget API config.

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/comp-3596


dgwatkins:
@OnTheGoSystems/compatibility please review

strategio (approved):
+1 @dgwatkins

Perhaps you can start a CC pipeline only for Cornerstone (once the config file is propagated). It might not fix everything but maybe some.

PR #197, opened 2021-04-23
Comp 3618

izzi-hassan:
Adding Support for Oxygen Builder in the first commit. Please let me know if there's anything else that needs to happen to override local repository.

Second one is just sorting the Config Index, feel free to cherry pick the commit if there are any issue with that @dgwatkins

### DIFF
--- a/config-index.xml
+++ b/config-index.xml
@@ -77,6 +77,7 @@
 		<item id="click-to-chat-for-whatsapp" override_local="true">Click to Chat</item>
 		<item id="seo-by-rank-math" override_local="true">Rank Math SEO</item>
 		<item id="stackable-ultimate-gutenberg-blocks" override_local="true">Stackable - Gutenberg Blocks</item>
+		<item id="oxygen" override_local="true">Oxygen Builder</item>
 	</plugins>
 	<themes>
 		<item id="avada" override_local="true">Avada</item>
--- a/config-index.xml
+++ b/config-index.xml
@@ -1,31 +1,52 @@
 <configurations>
 	<plugins>
 		<item id="10bit-woocommerce-gateway-yaadpay" override_local="true">10bit WooCommerce Gateway Yaadpay</item>
+		<item id="advanced-gutenberg" override_local="true">Advanced Gutenberg</item>
+		<item id="advanced-gutenberg" override_local="true">PublishPress Blocks</item>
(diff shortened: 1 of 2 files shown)


izzi-hassan:
@dgwatkins @amElnagdy please review!

izzi-hassan:
I have added some more XML config for the missing translatable widgets.

We are still missing support for the image, video and code block elements. I'll discuss this in MM.

strategio:
Just one doubt @izzi-hassan: What is exactly the `ct_options` encoding? Is it handled by Oxygen Builder somewhere? Do you manage to set this encoding in WPML > Settings > Custom XML Config?

izzi-hassan:
@strategio I'm improving this with a more oxygen-specific encoding name. I'm also discussing this in MM with David.

The code for handling this encoding is slightly different to how we handle regular urlencoded json:
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-5606#focus=Comments-102-461759.0-0

dgwatkins (commented):
@izzi-hassan It looks good :+1:

Can you just double check that users without the `oxygen_ct_options` encoding won't have issues? Thanks!

strategio:
Thanks @izzi-hassan! Indeed it's clearer to have a prefixed encoding name if it's a custom one.

And do you manage to set this encoding in WPML > Settings > Custom XML Config? I am asking because I am not sure of the rules we have in our `xsd` file. But after thinking twice, I believe it will work because now we are also validating our PR with the `xsd`.

LGTM +1

strategio (approved):

izzi-hassan:
Yes, there are no issues with that custom encoding name being added to the WPML -> Settings -> Custom XML Configuration box.

Without the filters to handle this new encoding, we are facing the bug reported by our users, where we corrupt the translated json by running `htmlspecialchars` on it in `WPML\LIB\WP\Gutenberg\WPML_PB_Update_Shortcodes_In_Content::filter_attribute_translation `.

This is reported here:
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-5606

Adding this encoding to the config does not change the behavior or throw any errors or exceptions even without the filters @dgwatkins

Please merge, and then I'll give everything one last test before publishing an erratum!

dgwatkins (approved):
according to XSD:
- attribute encoding can be any string
- custom field encoding is restricted to `[0-9a-z,]*`

so :+1: @izzi-hassan

PR #198, opened 2021-04-27
Updates on "et_pb_slide"

diiegopereira:
Added "background_image_tablet" and "background_image_phone" support


dgwatkins (approved):
I am always worried if Extra theme will need the same changes.

But it has been lagging for some time.

I will approve here and take a note about updating extra some time soon.

PR #210, opened 2021-05-26
Exclude CF7 from "Translate Everything"

strategio:
The fist commit is just to copy the existing local config and make it the source of truth.


dgwatkins (approved):
:+1: @strategio

PR #224, opened 2021-07-05
SiteOrigin: set googlemap consent strings as translatable

dgwatkins:


strategio:
@dgwatkins, did you check that it does not break the whole config if it runs with an older version of WPML (without your fix)?

dgwatkins:
> @dgwatkins, did you check that it does not break the whole config if it runs with an older version of WPML (without your fix)?

@strategio it doesn't break .. its just ignored.

strategio (approved):
LGTM @dgwatkins +1

PR #238, opened 2021-09-29
Adds et_pb_menu shortcode

ofwald:
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-5874

We need it so the style applied to the original is applied to the translation. There are no attributes to translate.


dgwatkins (approved):

dgwatkins:
Merged and sync'd. .. thanks!

PR #240, opened 2021-10-14
Corrected icon aria-label attribute for icon block

bfintal:


strategio:
Hi @bfintal, did you have a chance to look into this again?

bfintal:
Hey @strategio, apologies for just getting back now, and thanks for the heads up. Yes, it really should have been `@aria-label` from the start, but we incorrectly added in `@alt` :) So it should be okay to replace it completely.

strategio:
Thanks for the confirmation @bfintal!

I've merged your fix in our `master` branch and synchronized it to our CDN (it should propagate now).

bfintal:
Thanks @strategio !

PR #246, opened 2021-12-10
Update wpml-config.xml

sulmanpucit:
Seeing PHP notices like this -> https://share.getcloudapp.com/E0uKxpJw when translation is added for layout section. The term is also missing for the translated items.

I believe fusion_tb_category should not be translatable just like element_category

### DIFF
--- a/avada/wpml-config.xml
+++ b/avada/wpml-config.xml
@@ -53,7 +53,7 @@
         <taxonomy translate="1">portfolio_tags</taxonomy>
         <taxonomy translate="1">themefusion_es_groups</taxonomy>
         <taxonomy translate="1">slide-page</taxonomy>
-        <taxonomy translate="1">fusion_tb_category</taxonomy>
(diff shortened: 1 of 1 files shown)


dgwatkins:
Hi @sulmanpucit
Thanks for the contribution. Let me run a few tests to make sure this is the best solution and I will get back to you.

dgwatkins:
@sulmanpucit I tried to translate a section and I did not get the same error messages as you.

Can you please open a support ticket in our forum so that we can analyze your problem better?

https://wpml.org/forums/forum/english-support/

dgwatkins:
@sulmanpucit Meanwhile you can still have this configuration in your setup by adding the following XML to WPML > Settings > Custom XML Configuration:
```
<wpml-config>
<taxonomies>
<taxonomy translate="0">fusion_tb_category</taxonomy>
</taxonomies>
</wpml-config>
```

I will close here for now and re-open when we have more information.
Thanks again

PR #254, opened 2022-01-11
[wpmldev-275]Make WP template and template part post types translatable

brucepearson:
Make WP template and template part post types translatable - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmldev-275


dgwatkins (approved):
:+1: @brucepearson 

I sync'd the change to S3

PR #258, opened 2022-02-10
Revert copy of category thumbnail_id

dgwatkins:
Revert this change because we have code handling this field already

https://onthegosystems.myjetbrains.com/youtrack/issue/wcml-3896


strategio (approved):
+1 @dgwatkins

PR #259, opened 2022-02-15
Add support to translate Enfold select field options

andrescifuentesr:
- Add support to translate the select field options available when using the Contact module


dgwatkins (approved):

dgwatkins:
Thanks @andrescifuentesr 

Its sync'd to S3 now

PR #260, opened 2022-02-21
[Avada] Added support for some form fields

diiegopereira:
[Avada] Added support for "fusion_form_text" and "fusion_form_phone_number" shortcodes


dgwatkins:
Thanks @diiegopereira 
Sync'd to S3

PR #268, opened 2022-04-29
Update wpml-config.xml for Mix and Match 2

helgatheviking:
Copy new meta fields that were added in MNM 2.0. Follow up to compat code: https://git.onthegosystems.com/glue-plugins/wpml/woocommerce-multilingual/-/merge_requests/1674


dgwatkins (approved):
Thanks @helgatheviking

PR #269, opened 2022-05-09
Fix from compsupp-5363

diiegopereira:


dgwatkins (approved):

dgwatkins:
Hi @diiegopereira 
I merged the branch but I was wondering if we wont need the same change in `js_composer` and `js_composer-dt-the7`?

diiegopereira:
You're right @dgwatkins, we got another report for js_composer on compsupp-6148.

PR #272, opened 2022-05-10
Removed the encoding from vc_single_image

diiegopereira:
It seems that the encoding on link (vc_single_image) was causing some issues. See:
- https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-6148
- https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-6209

### DIFF
--- a/js_composer/wpml-config.xml
+++ b/js_composer/wpml-config.xml
@@ -121,7 +121,7 @@
             <attributes>
                 <attribute encoding="allow_html_tags">title</attribute>
                 <attribute type="media-ids">image</attribute>
-                <attribute encoding="vc_link" type="link">link</attribute>
+                <attribute>link</attribute>
             </attributes>
         </shortcode>
         <shortcode>


dgwatkins:
Hi @diiegopereira 

1) I understand about removing `encoding="vc_link"`. The link is not encoded so that's fine.
2) But I don't see a reason to remove `type="link"` because then internal links wont be auto-translated
3) I suppose we need this change in the other two versions of js_composer that we have

strategio:
@diiegopereira, @dgwatkins, what's the status of this PR?

diiegopereira:
@dgwatkins @strategio this PR is a bit old and we haven't had any new reports since. So I think we should re-check the correct XML code for this widget.

I created this ticket to re-test the widget: https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-6451 - I will create a new PR after that.

PR #278, opened 2022-06-03
WPBakery Page Builder - Set post custom CSS to copy

5um17:
WPBakery Page Builder has post specific custom CSS option, which is not copied or translated on post translation.
The value is stored in post meta with key `_wpb_post_custom_css`.

Ref:- https://wpml.org/forums/topic/the-background-color-at-the-top-of-the-main-page-is-not-visible/

### DIFF
--- a/js_composer/wpml-config.xml
+++ b/js_composer/wpml-config.xml
@@ -7,6 +7,7 @@
     <custom-fields>
         <custom-field action="ignore">_vc_post_settings</custom-field>
         <custom-field action="copy">_wpb_shortcodes_custom_css</custom-field>
+        <custom-field action="copy">_wpb_post_custom_css</custom-field>
     </custom-fields>
     <shortcodes>
         <shortcode>


strategio:
@5um17, unfortunately we cannot do this as a general rule because it would overwrite a possible custom CSS in the translation if it was edited manually.

We have some code about that in `WPML\Compatibility\WPBakery\Styles`. It could be a regression, could you please check?

5um17:
Thanks @strategio! I understand. I have checked the code in `WPML\Compatibility\WPBakery\Styles` that is what is needed. But we are handling only one key `_wpb_shortcodes_custom_css` but the problem is here with `_wpb_post_custom_css` so not a regression IMHO.

Do you want me to open a ticket for this?

strategio:
Yes, escalate please @5um17 :)

strategio:
@5um17, can we close this PR?

5um17:
yes sure.
FTR, It is fixed in wpmlpb-272

PR #280, opened 2022-06-28
Updated support for hotspot widget (compsupp-6303)

diiegopereira:
See compsupp-6303

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -254,7 +254,8 @@
         <widget name="hotspot">
             <fields-in-item items_of="hotspot">
                 <field type="Hotspot: Label">hotspot_label</field>
-                <field type="Hotspot: URL" editor_type="LINK">hotspot_link>url</field>
+        	<field type="Hotspot: URL" key_of="hotspot_link">url</field>
+		<field type="Hotspot: URL" key_of="__dynamic__">hotspot_link</field>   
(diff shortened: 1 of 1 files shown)


elementor/wpml-config.xml
```diff
@@ -254,7 +254,8 @@
         <widget name="hotspot">
             <fields-in-item items_of="hotspot">
                 <field type="Hotspot: Label">hotspot_label</field>
-                <field type="Hotspot: URL" editor_type="LINK">hotspot_link>url</field>
+        	<field type="Hotspot: URL" key_of="hotspot_link">url</field>
+		<field type="Hotspot: URL" key_of="__dynamic__">hotspot_link</field>
```
strategio, line 258: @diiegopereira, are you sure we need that (the `__dynamic__` part)?

We have some integration code in https://git.onthegosystems.com/glue-plugins/wpml/wpml-page-builders/-/blob/2.0.4/classes/Integrations/Elementor/Config/DynamicElements/Hotspot.php#L21.

diiegopereira: @strategio in this case I think we can remove the __dynamic__ part. I added it just to cover both possibilities, but we already have this covered in our code.

strategio (approved):
+1 @diiegopereira

strategio:
Deployed.

PR #285, opened 2022-08-05
WPML config for dt-the7 10.12.0

gerasimovdaniel:


dt-the7/wpml-config.xml
```diff
@@ -92,6 +92,19 @@
         <field type="Button Text" editor_type="LINE">read_more_button_text</field>
       </fields>
     </widget>
+    <widget name="the7-woocommerce-cart-preview">
+      <fields>
+        <field type="Widget Title" editor_type="LINE">widget_title_text</field>
+        <field type="Subtotal Title" editor_type="LINE">subtotal_text</field>
+        <field type="Button Name" editor_type="LINE">shop_button_text</field>
+        <field type="Link" editor_type="LINK" key_of="shop_button_link">url</field>
```
strategio, line 100: @gerasimovdaniel, we have a new way to define this data by path:

```xml
<field type="Link" editor_type="LINK">shop_button_link>url</field>
```

We are trying to move away from this legacy `key_of` attribute (bad design decision).

[See an example here](https://github.com/OnTheGoSystems/wpml-config/blob/master/elementor/wpml-config.xml#L26)

Note that for now, the path is supported only for `fields` sub-items.

dt-the7/wpml-config.xml
```diff
@@ -92,6 +92,19 @@
...
+        <field type="Button Name" editor_type="LINE">view_cart_button_text</field>
+        <field type="Checkout Text" editor_type="LINE">checkout_button_text</field>
+        <field type="Empty Cart Text" editor_type="LINE">empty_cart_text</field>
+        <field type="Button Text" editor_type="LINE">empty_cart_button_shop_text</field>
+        <field type="Link" editor_type="LINK" key_of="empty_cart_button_link">url</field>
```
strategio, line 105: Same as for my previous comment, could you please use the path way:

```xml
<field type="Link" editor_type="LINK">empty_cart_button_link>url</field>
```

strategio (changes requested):
Thanks @gerasimovdaniel for your contribution!

I've left a request for 2 small changes.

And I see that [our documentation](https://wpml.org/documentation/support/language-configuration-files/how-to-register-page-builder-widgets-for-translation/) should be updated accordingly (I'll take care of that).

strategio:
Hi @gerasimovdaniel!

Is that PR still valid? Do you need help to address my comments?

Thanks!

gerasimovdaniel:
Hi @strategio!

Thanks for review!
Changes have been made.

Can you recommend any tools to generate Elementor part of config? I've made my own, but this `>` in element value seems to be invalid XML syntax and turned to be encoded.

Thank you!

dt-the7/wpml-config.xml
gerasimovdaniel: Done

dt-the7/wpml-config.xml
gerasimovdaniel: Done

strategio:
Thanks @gerasimovdaniel!

However, we prefer not to update existing configurations for now. The reason if that the string name signature is not exactly the same and it could cause minor issues for people using the Classic Translation Editor that has a different approach for the translation memory.

It should not be a big issue, but we prefer to stay on the safe side.

And so I've created another PR where I took your initial changes and applied my comment suggestions directly.

Could you please give a look at https://github.com/OnTheGoSystems/wpml-config/pull/299 and confirm this config is working for you?

PR #289, opened 2022-08-18
Added support for shortcodes from the Bloom plugin

diiegopereira:
See compsupp-6339

PR #290, opened 2022-08-18
Added support for shortcodes from the Bloom plugin

strategio:
See compsupp-6339


strategio:
Already merged in https://github.com/OnTheGoSystems/wpml-config/pull/289.

PR #291, opened 2022-08-29
compsupp-6331 - Added OptinMonster settings

diiegopereira:
See https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-6331

It's ready for publish/deploy

### DIFF
--- a/optinmonster/wpml-config.xml
+++ b/optinmonster/wpml-config.xml
@@ -1,5 +1,5 @@
 <wpml-config>
-  <custom-types>
- 		<custom-type translate="0">omapi</custom-type>
+	<custom-types>
+		<custom-type translate="0">omapi</custom-type>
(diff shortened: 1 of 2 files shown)


optinmonster/wpml-config.xml
```diff
@@ -0,0 +1,5 @@
+<wpml-config>
+  <custom-types>
```
dgwatkins, line 2: @diiegopereira let's be consistent and use either spaces or tabs, I would go for tabs.

dgwatkins (changes requested):

optinmonster/wpml-config.xml
diiegopereira: @dgwatkins thanks! I fixed it in the last commit. Please check.

PR #296, opened 2022-10-19
[wpmldev-1014] Adding core/list-item block type to gutenberg after ch…

gregWierzba:
…anges in WP 6.1


dgwatkins:
@gregWierzba can you make sure that the `wpml-config.xml` file in core is also updated?

gregWierzba:
> 

yes, @dgwatkins I will apply that change to core

dgwatkins (approved):
looks good @gregWierzba

PR #307, opened 2023-02-13
Added config for Hummingbird

diiegopereira:
See compsupp-6626

This issue is affecting ATE, so I set override_local="true"


dgwatkins:
deployed @diiegopereira

PR #312, opened 2023-03-01
[Elementor] Translate "price" from Price list widget

andrescifuentesr:
Making the Price field translatable for Price list widget
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-6636

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -341,8 +341,7 @@
             <fields-in-item items_of="price_list">
                 <field type="Pricing list: title">title</field>
                 <field type="Pricing list: description">item_description</field>
-                <field type="Pricing list: price">price</field>
-                <field type="Pricing list: link">link>url</field>        
+                <field type="Pricing list: link" editor_type="LINK">link>url</field>
             </fields-in-item>
         </widget>	    
     </elementor-widgets>
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -337,5 +337,13 @@
                 <field type="Countdown: message after expire">message_after_expire</field>
             </fields>
         </widget>
+        <widget name="price-list">
+            <fields-in-item items_of="price_list">
+                <field type="Pricing list: title">title</field>
+                <field type="Pricing list: description">item_description</field>
+                <field type="Pricing list: price">price</field>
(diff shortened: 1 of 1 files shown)


dgwatkins:
@andrescifuentesr I don't think we shoud make the price translatable for everyone:

- ATE doesn't know how to translate a price
- Most customers wont want to translate prices anyway.

If a certain customer needs, he can set it in his local XML.

What do you think?

andrescifuentesr:
@dgwatkins I had the same doubts about setting this field to translate or not. The initial request was done because the user was not adding just a number but also currencies and weight, ie 5€/kg.
So, if you just add a number, that would be OK as the value will be copied directly, however, if you add a currency, you may want to translate it.
Anyway, we can leave it as it is and collect more reports, if it makes sense.

dgwatkins:
Thanks @andrescifuentesr 

I will close here then and we can follow up on the ticket.

dgwatkins:
Oops, sorry @andrescifuentesr 

I realize now that this PR has more than just the price.

I think you should remove the price from this pull request and then we can merge.
Makes sense?

elementor/wpml-config.xml
```diff
@@ -337,5 +337,13 @@
                 <field type="Countdown: message after expire">message_after_expire</field>
             </fields>
         </widget>
+        <widget name="price-list">
+            <fields-in-item items_of="price_list">
+                <field type="Pricing list: title">title</field>
+                <field type="Pricing list: description">item_description</field>
+                <field type="Pricing list: price">price</field>
```
dgwatkins, line 344: Remove the above line @andrescifuentesr

elementor/wpml-config.xml
```diff
@@ -337,5 +337,13 @@
...
+                <field type="Pricing list: link">link>url</field>
```
dgwatkins, line 345: And lets add the editor_type here:
```suggestion
                <field type="Pricing list: link"editor_type="LINK">link>url</field>        
```

dgwatkins (changes requested):

andrescifuentesr:
Thanks for the help, David.
I think we are good to go.

dgwatkins (approved):

dgwatkins:
fixed,  merged and deployed @andrescifuentesr

PR #319, opened 2023-05-11
feat: make the excerpt block link translatable

decodekult:
Ref: wpmlpb-325


strategio (approved):
+1 @decodekult

strategio:
Merged an deployed @decodekult

PR #320, opened 2023-05-31
Updates for ultimate-addons-for-gutenberg (now Spectra)

diiegopereira:
See compsupp-6801

### DIFF
--- a/config-index.xml
+++ b/config-index.xml
@@ -56,6 +56,7 @@
 		<item id="stackable-ultimate-gutenberg-blocks" override_local="true">Stackable - Gutenberg Blocks</item>
 		<item id="subscriptio">Subscriptio</item>
 		<item id="the-events-calendar" override_local="true">The Events Calendar</item>
+		<item id="ultimate-addons-for-gutenberg" override_local="true">Spectra</item>
 		<item id="ultimate-addons-for-gutenberg" override_local="true">Ultimate Addons for Gutenberg</item>
 		<item id="ultimate-member" override_local="true">Ultimate Member</item>
 		<item id="uncode-js_composer" override_local="true">Uncode Page Builder (Visual Composer)</item>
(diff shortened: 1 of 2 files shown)


config-index.xml
```diff
@@ -56,6 +56,7 @@
 		<item id="stackable-ultimate-gutenberg-blocks" override_local="true">Stackable - Gutenberg Blocks</item>
 		<item id="subscriptio">Subscriptio</item>
 		<item id="the-events-calendar" override_local="true">The Events Calendar</item>
+		<item id="ultimate-addons-for-gutenberg" override_local="true">Spectra</item>
```
strategio, line 59: @diiegopereira, I saw your comment in `compsupp-6801`, that the Spectra plugin has a local WPML config file.

Beware that if we override the local file, we may loose some local configurations if it was not synced on the remote one.

Can you make sure the local and remote config files are in sync?

diiegopereira: @strategio it seems that we have several differences between our remote file and the one included in the plugin: https://www.diffchecker.com/ArriH1zI/

I was trying to manually update the remote file by adding the differences to it. But I found some parts that can generate conflicts. E.g. in the remote file, the string is registered using xPath, but in the plugin file, it is registered using "key" (or a different xPath that belongs to the same string).

What would be the correct way to sync both files? Should we use what's included in the plugin, which looks more up-to-date? Is there a problem with a string being registered twice (key + xPath, or 2 different xPath for the same string)?

strategio: @diiegopereira, let me share some elements:

1. String IDs (i.e. strings names) are based on the block name and the text content, we make a hash of that (see [here](https://git.onthegosystems.com/glue-plugins/wpml/wpml-page-builders/-/blob/master/classes/Integrations/Gutenberg/strings-in-block/class-base.php#L133)). In short, strings are unique to block type and content. If we have twice the same content inside a block, we'll register only 1 string. So we don't risk to have duplicated strings.

2. If you see that the plugin's local XML config looks more up-to-date than the remote one, it means that the author is maintaining his config (even not fully). At least, he probably does that more than us.

With all that said, I think we have 2 options:

1. Take real ownership of the config: First, we need to merge the local config into the remote one (as you started to do), and then inform the author that the remote configuration will now take precedence. If he wants to change/add blocks in the future, he will need to open a PR here.

2. Let the author take care of his config, and inform him about the issue we found (and perhaps ask him for a campaign of re-testing on his plugin).

I'll let you check what could be the best option.

diiegopereira: Thanks @strategio! I think we should contact the author and discuss the best option. Let's continue on youtrack for now.

PR #323, opened 2023-07-06
Enable automatic translation of Contact Form 7 forms

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcf7-55/Contact-Form-7-post-types-should-be-enabled-for-Automatic-Translation-if-TE-is-enabled.


strategio (commented):
@dgwatkins, did you review your PR? I don't understand it, you are removing all CF7ML config.

dgwatkins:
> @dgwatkins, did you review your PR? I don't understand it, you are removing all CF7ML config.

Yes @strategio, We only added this XML to disable automatic translation.

If we want it automatic now, we can just remove the whole thing.

dgwatkins:
In other words, the task here is to remove `automatic="0"`. Removing that would leave the file identical to the one shipped with the plugin so we no longer need this remote file.

Keeping in mind that we normally like to keep remove configs to a minimum, I took the option of removing it.

strategio (approved):
Thanks for the clarification @dgwatkins +1

dgwatkins:
merged and deployed

PR #344, opened 2023-12-12
Updated js_composer with icon and pricing table links

diiegopereira:
See compsupp-7086

### DIFF
--- a/js_composer-dt-the7/wpml-config.xml
+++ b/js_composer-dt-the7/wpml-config.xml
@@ -397,6 +397,13 @@
             <tag>vc_icon</tag>
             <attributes>
                 <attribute>title</attribute>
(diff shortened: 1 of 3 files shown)


strategio:
@diiegopereira, we are trying to synchronize the `js_composer` configuration (WP Bakery) for when it's also embedded in themes. See https://github.com/OnTheGoSystems/wpml-config/commit/551e61821f84231afdfc043137638669e48d9794.

If the new shortcode is generic, could you also include it in the other configuration files (`js_composer-dt-the7/wpml-config.xml` and `uncode-js_composer/wpml-config.xml`).

diiegopereira:
@strategio sure, I added it to both files now.

dgwatkins (approved):

PR #347, opened 2024-01-04
Set labels for ATE Gen3

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-419

### DIFF
--- a/Divi/wpml-config.xml
+++ b/Divi/wpml-config.xml
@@ -103,23 +103,23 @@
             </attributes>
         </shortcode>
         <shortcode>
-            <tag>et_pb_cta</tag>
+            <tag label="Call To Action: Content">et_pb_cta</tag>
             <attributes>
-                <attribute>title</attribute>
-                <attribute>admin_label</attribute>
-                <attribute>button_text</attribute>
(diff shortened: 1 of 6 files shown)


dgwatkins:
I believe we can merge this already because we are just updating the labels.

@OnTheGoSystems/compatibility

dgwatkins:
Note that we don't need specific labels for every element because we try to put together the label automatically. We only set a specific label when the automatic label doesn't match the requirements.

strategio (commented):
@dgwatkins, I quickly scanned the changes and it looks good. The only question is about the order of the items in the config: Our current config does not respect a clear order (alphabetical or other), and in your changes you also changed the order of some items, but still it's not alphabetical. Did you opted for another specific order?

dgwatkins:
No @strategio

As far as I can see, the order of the XML elements has nothing to do with it.
ATE receives string in the order they are registered in (string_location field).

I did change the order of divi shortcode definitions to confirm this.

I left it changed because it makes more sense to have the titles at the top and the rest below.

strategio (approved):
+1 @dgwatkins

PR #364, opened 2024-02-28
WPBakery - Adding support for SEO feature

andrescifuentesr:
WPBakery - Adding support for SEO feature
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7221


strategio:
@dgwatkins, I think you should review that. Also to check if it has something to do with the new groupings/sortings in the ATE.

PR #365, opened 2024-03-04
[acfml-798] Options pages should not be translatable.

decodekult:
Instead, we will offer to translate their labels.

https://onthegosystems.myjetbrains.com/youtrack/issue/acfml-798/


dgwatkins (approved):

dgwatkins:
Merged and deployed @decodekult

PR #367, opened 2024-03-12
Enfold - Enabling Media Translation for column background images

andrescifuentesr:
- Allowing that images added as background in columns can be translated with Media Translation
- https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7248


strategio (approved):
+1 @andrescifuentesr

strategio:
Merged and deployed!

PR #368, opened 2024-03-15
[Elementor] Adding support for Lottie File translation

andrescifuentesr:
- Adding support for Lottie File translation, caption and links
- https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7263

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -384,5 +384,13 @@
                 <field type="Loop Grid: Nothing found message">nothing_found_message_text</field>
             </fields>
         </widget>
+        <widget name="lottie">
+            <fields>
+                <field type="Lottie: Media File" editor_type="LINK">source_json>url</field>
+                <field type="Lottie: External URL" editor_type="LINK">source_external_url>url</field>
+                <field type="Lottie: Caption" editor_type="LINE">caption</field>
+                <field type="Lottie: Link" editor_type="LINK">custom_link>url</field>
+            </fields>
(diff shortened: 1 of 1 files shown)


elementor/wpml-config.xml
```diff
@@ -384,5 +384,13 @@
                 <field type="Loop Grid: Nothing found message">nothing_found_message_text</field>
             </fields>
         </widget>
+        <widget name="lottie">
+            <fields>
+                <field type="Lottie: Media File" editor_type="LINE">source_json>url</field>
```
strategio, line 389: Shouldn't we use `LINK` for URL @andrescifuentesr?

@dgwatkins, since you are making important changes for the new ATE, I think you should review that too.

andrescifuentesr: Thanks for the heads up @strategio. I'll add the change.

dgwatkins: Yes, we should have type="LINK"

dgwatkins: The way things are at the momenet, external links will be converted to "LINE". Better leave it as a "LINK" anyway, in case we decide to treat it differently.

dgwatkins (approved):
:+1: @andrescifuentesr

dgwatkins:
I used to squash using the UI, but Pierre asked us not to do so.

Can you squash @andrescifuentesr ?

dgwatkins:
Squashed, merged and deployed

PR #371, opened 2024-04-04
[The7] Added config file with override_local false

andrescifuentesr:
- Setting override_local as false
- Keeping js_composer-dt-the7 as true for consistency.  
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7258/


dgwatkins:
@andrescifuentesr do you know if "The7" author has been keeping his XML up to date? I am asking because as soon as I merge this, all "The7" users will stop receiving our XML and start using their local XML.

FYI: @gerasimovdaniel

andrescifuentesr:
@gerasimovdaniel, do you think is better to wait until The7 new version is released and give time for your users to update and then remove it? Or maybe you just prefer to keep maintaining our repository as you did this morning?
Let us know your thoughts.

gerasimovdaniel:
@ andrescifuentesr, I prefer to keep the XML in The7. We are keeping it up to date quite rigorously. Let's wait until the new version of The7 is released. Should I post here when it comes?

andrescifuentesr:
@gerasimovdaniel, sure. Please let us know once the new version is released so we can go further.
Let's hold this on @dgwatkins.

gerasimovdaniel:
@dgwatkins The7 new version is released.

diiegopereira:
Implementing this change may cause issues for users with older versions of The7 theme, where the new XML file has not yet been implemented. The WPML integration will suddenly stop working.

The solution in this case would be to keep both files synchronized (local and remote) for a while, until most users have updated the theme.

I talked to Yana via email and we decided to wait around a year to do this modification.

dgwatkins:
@andrescifuentesr I think we can remove the comments, they are not adding value.

andrescifuentesr:
Thanks for the heads up @dgwatkins, something went wrong on my local branch. Sorry for the noise.

dgwatkins (approved):

dgwatkins:
@andrescifuentesr Let me know when we are safe to deploy this.

andrescifuentesr:
Yes @dgwatkins, we are good to go with this, thanks!

PR #373, opened 2024-04-08
[Kadence Block] - Setting override_local as false

andrescifuentesr:
- Setting override_local as false 
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7019


dgwatkins (changes requested):
@andrescifuentesr disabling override and removing the whole XML is the same.

If the author is ready to take over, lets just remove the whole XML.

PR #376, opened 2024-05-17
[Spectra] Adding support for Infobox and Buttons blocks

andrescifuentesr:
- Infobox: adding more robust support for translating the title.
- Button: making link translatable
https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7376 https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7380

### DIFF
--- a/ultimate-addons-for-gutenberg/wpml-config.xml
+++ b/ultimate-addons-for-gutenberg/wpml-config.xml
@@ -18,12 +18,11 @@
 			<key name="ctaText"/>
 		</gutenberg-block>
 		<gutenberg-block type="uagb/info-box" translate="1">
-			<xpath>//span[@class="uagb-ifb-title-prefix"]</xpath>
-			<xpath>//*[@class="uagb-ifb-title"]</xpath>
-			<xpath>//a/@href</xpath>
-			<xpath>//*[self::h1 or self::h2 or self::h3 or self::h4 or self::h5 or self::h6]</xpath>
-			<xpath>//p[@class="uagb-ifb-desc"]</xpath>
-			<xpath>//span[contains(@class, 'uagb-inline-editing')]</xpath>
+			<xpath label="Title Prefix">//span[@class="uagb-ifb-title-prefix"]</xpath>
+			<xpath label="Title">//*[self::h1 or self::h2 or self::h3 or self::h4 or self::h5 or self::h6 or self::p]</xpath>
+			<xpath label="Description">//p[@class="uagb-ifb-desc"]</xpath>
+			<xpath label="Button Text">//span[contains(@class, 'uagb-inline-editing')]</xpath>
+			<xpath label="Button Link" type="link">//a/@href</xpath>
 		</gutenberg-block>
 		<gutenberg-block type="uagb/testimonial" translate="1">
 			<xpath>//div[@class="uagb-tm__desc"]</xpath>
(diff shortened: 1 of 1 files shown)


ultimate-addons-for-gutenberg/wpml-config.xml
```diff
@@ -73,6 +74,7 @@
 		<gutenberg-block type="uagb/buttons-child" translate="1">
 			<xpath>//*[@class="uagb-button__link"]</xpath>
 			<xpath>//*[@class="uagb-buttons-repeater"]/@href</xpath>
+			<xpath>//a/@href</xpath>
```
dgwatkins, line 77: @andrescifuentesr wouldn't this have to be a `type="link"`?

I see another one missing in the `info-box`, please review if there are any more in this file.

ultimate-addons-for-gutenberg/wpml-config.xml
```diff
@@ -19,6 +19,7 @@
 		</gutenberg-block>
 		<gutenberg-block type="uagb/info-box" translate="1">
 			<xpath>//span[@class="uagb-ifb-title-prefix"]</xpath>
+			<xpath>//*[@class="uagb-ifb-title"]</xpath>
```
dgwatkins, line 22: @andrescifuentesr Looking at the source code, it looks like the title will be caught by the pattern 2 lines below
`//*[self::h1 or self::h2 or self::h3 or self::h4 or self::h5 or self::h6]`

If that pattern was not enough and you this one works better, is it possible to replace it instead?


In other words, we now have 2 patterns matching the same string and I'm not sure if thats a good idea.

dgwatkins (dismissed):
Would it make sense to add labels here @andrescifuentesr?
The order is also important, we should use the same order as we see in the editor
```
      <xpath label="Title Prefix">//span[@class="uagb-ifb-title-prefix"]</xpath>
      <xpath label="Title">//*[@class="uagb-ifb-title"]</xpath>
      <xpath label="Description">//p[@class="uagb-ifb-desc"]</xpath>
      <xpath label="Button Text">//span[contains(@class, 'uagb-inline-editing')]</xpath>
      <xpath label="Button Link" type="link">//a/@href</xpath>
```

ultimate-addons-for-gutenberg/wpml-config.xml
andrescifuentesr: @dgwatkins, you are right. In that case we would need to modify this pattern instead and add a p tag that was missing:
`<xpath>//*[self::h1 or self::h2 or self::h3 or self::h4 or self::h5 or self::h6 or self::p]</xpath>`
In that way we avoid possible issues with legacy

dgwatkins: But that way you will catch any <p> tag, I don't think we want that @andrescifuentesr 

I would remove the above pattern and keep the new one. It will also catch the h1..6 because it has a wildcard `*`

andrescifuentesr: Thanks for your inputs @dgwatkins. Done!

dgwatkins (approved):
Nice :+1: @andrescifuentesr

dgwatkins:
merged and deployed @andrescifuentesr

PR #384, opened 2024-06-13
Updated PRO theme settings

diiegopereira:
In wpmlpb-283, we've made some modifications to the Cornerstone XML to add support for classic widgets.

The PRO theme has Cornerstone bundled, so we also need to update its XML.

See compsupp-7412 for more details

### DIFF
--- a/pro/wpml-config.xml
+++ b/pro/wpml-config.xml
@@ -285,4 +285,196 @@
       </attributes>
     </shortcode>
   </shortcodes>
+  <cornerstone-widgets>
+    <widget name="classic:alert">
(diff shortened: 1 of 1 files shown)


strategio (approved):
I guess you've just copied the config @diiegopereira, right?

If yes, that's all good to me +1

diiegopereira:
Yes, I copied from the cornerstone XML: https://github.com/OnTheGoSystems/wpml-config/blob/master/cornerstone/wpml-config.xml

strategio:
Merged and deployed!

PR #385, opened 2024-06-17
Updated configs for Spectra

diiegopereira:
See compsupp-7419

### DIFF
--- a/ultimate-addons-for-gutenberg/wpml-config.xml
+++ b/ultimate-addons-for-gutenberg/wpml-config.xml
@@ -54,8 +54,8 @@
 			<xpath>//p[@class="uagb-cta__desc"]</xpath>
 			<xpath type="link">//a[contains(@class,"uagb-cta__block-link")]/@href</xpath>
 			<xpath>//span[@class="uagb-cta__link-content-inner"]/span</xpath>
-			<xpath>//*[@class="uagb-cta__buttons"]</xpath>
-      			<xpath>//a[@class="uagb-cta__button-link-wrapper"]</xpath>
+			<xpath>//*[@class="uagb-cta__buttons"]/a</xpath>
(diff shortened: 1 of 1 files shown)


ultimate-addons-for-gutenberg/wpml-config.xml
```diff
@@ -54,6 +54,8 @@
 			<xpath>//p[@class="uagb-cta__desc"]</xpath>
 			<xpath type="link">//a[contains(@class,"uagb-cta__block-link")]/@href</xpath>
 			<xpath>//span[@class="uagb-cta__link-content-inner"]/span</xpath>
+			<xpath>//*[@class="uagb-cta__buttons"]</xpath>
+      			<xpath>//a[@class="uagb-cta__button-link-wrapper"]</xpath>
```
dgwatkins, line 58: These 2 lines are both targetting the same element @diiegopereira 

According to the ticket, the structure is more or less like this:
```
<div class="uagb-cta__buttons"><a href="#" class="uagb-cta__button-link-wrapper">Read More</a></div>
```

One way is to select `uagb-cta__buttons` and let ATE parse the links (links are translated inline):
```suggestion
			<xpath>//*[@class="uagb-cta__buttons"]</xpath>
```
Another way would be to target link and text independently, but I am not sure what you need:
```suggestion
			<xpath>//*[@class="uagb-cta__buttons"]/a</xpath>
			<xpath type="link">//*[@class="uagb-cta__buttons"]/a/@href</xpath>
```

diiegopereira: Thanks David, I would go with the second option. I tested and it works.

dgwatkins (approved):

dgwatkins:
Merged and deployed @diiegopereira

PR #387, opened 2024-06-26
[compsupp-7433] Cornerstone - Adding support for new buttons

andrescifuentesr:
- Adding support for new buttons: allowing to translate label and. link
- https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7433/

### DIFF
--- a/cornerstone/wpml-config.xml
+++ b/cornerstone/wpml-config.xml
@@ -247,5 +247,11 @@
                 <field type="Testimonial: Content">testimonial_content</field>
             </fields>
         </widget>
+        <widget name="button">
+            <fields>
+                <field type="Button" editor_type="LINE">anchor_text_primary_content</field>
+                <field type="Button: Link URL" editor_type="LINK">anchor_href</field>
+            </fields>
+        </widget>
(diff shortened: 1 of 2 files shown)


cornerstone/wpml-config.xml
```diff
@@ -247,5 +247,11 @@
                 <field type="Testimonial: Content">testimonial_content</field>
             </fields>
         </widget>
+        <widget name="button">
+            <fields>
+                <field type="Button" editor_type="LINE">anchor_text_primary_content</field>
```
dgwatkins, line 252: @andrescifuentesr Please use the same prefix for both (so that ATE groups them together), something like this:
```suggestion
                <field type="Button: Button Text" editor_type="LINE">anchor_text_primary_content</field>
```
Same applies to Pro theme.

andrescifuentesr: It makes sense @dgwatkins, thanks for the suggestion.

dgwatkins (approved):

dgwatkins:
Squashed, merged and deployed @andrescifuentesr

PR #394, opened 2024-08-08
[Yoast] Adding support for multiple entries of Related keyphrases

andrescifuentesr:
- Adding support for multiple entries of Related keyphrases: Keyphrase and Synonyms.
-https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7515


strategio (commented):
@andrescifuentesr, is that still backward compatible? I mean, does it still work if you have a single entry?

andrescifuentesr:
Yes @strategio, it keeps working for a single entry.

PR #399, opened 2024-09-23
Updating Elementor XML - compsupp-7594

diiegopereira:
See https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7594

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -487,5 +487,10 @@
                 <field>icon_text</field>
             </fields-in-item>
         </widget>
+        <widget name="contact-buttons-var-4">
+            <fields-in-item items_of="contact_repeater">
+                <field>contact_tooltip</field>
+            </fields-in-item>
+        </widget>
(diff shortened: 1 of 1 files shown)


strategio (commented):
@diiegopereira, it looks good. Did you also find other variations of the name "contact-buttons-var-4" (maybe "contact-buttons-var-1"), perhaps we should also support it?

diiegopereira:
@strategio I checked Elementor code and found that they have at least 10 variations of this widget.

I believe we will have to test and generate the XML configuration for each of them, as they have different structures. For this, I created a new ticket (compsupp-7621).

Perhaps we can merge this PR, and then add the other configs in a new PR?

strategio (approved):
Thanks @diiegopereira +1

strategio:
That's merged @diiegopereira.

@dgwatkins, I could not connect to the server to deploy, could you please do it if you can?

dgwatkins:
Deployed @strategio

PR #401, opened 2024-09-26
[Spectra] Adding support for Container links

andrescifuentesr:
- Adding support for Container links
- https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7603/Spectra-UAGB-Container-link-translation-not-working

### DIFF
--- a/ultimate-addons-for-gutenberg/wpml-config.xml
+++ b/ultimate-addons-for-gutenberg/wpml-config.xml
@@ -126,5 +126,8 @@
 		<gutenberg-block type="uagb/forms" translate="1">
 			<xpath>//button/div</xpath>
 		</gutenberg-block>
+		<gutenberg-block type="uagb/container" translate="1">
+			<xpath label="Container link" type="link">//a/@href</xpath>
+		</gutenberg-block>		
 	</gutenberg-blocks>
 </wpml-config>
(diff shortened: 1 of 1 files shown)


ultimate-addons-for-gutenberg/wpml-config.xml
```diff
@@ -126,5 +126,8 @@
 		<gutenberg-block type="uagb/forms" translate="1">
 			<xpath>//button/div</xpath>
 		</gutenberg-block>
+		<gutenberg-block type="uagb/container" translate="1">
+			<xpath label="Container link" type="link">//a/@href</xpath>
```
dgwatkins, line 130: I would try to be more specific here because the container could have other content. I'm not 100% sure, but this could be safer:
```suggestion
			<xpath label="Container link" type="link">//a[contains(@class, "spectra-container-link-overlay")]/@href</xpath>
```
Its up to you, if you tested with other links inside the container and it works, feel free to ignore :-)

andrescifuentesr: I had the same initial doubt and I initially thought it was better for leave it less specific. However, your comment makes sense @dgwatkins.

dgwatkins (approved):

dgwatkins:
merged and deployed

PR #405, opened 2024-10-25
[Spectra] Making Info Box Block Title prefix translatable for other HTML tags

andrescifuentesr:
- Info Box Block Title prefix was translatable only for <span> tags
- https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7655


andrescifuentesr:
Info Box Block Title prefix was translatable only for 'span' tags

dgwatkins (approved):
👍 @andrescifuentesr

PR #411, opened 2024-11-27
[compsupp-7679] Uncode theme - Add translation to column links

nicolasviallet:
Uncode has an option to add a link to vc_column and make it clickable. 

Youtrack: https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7679

### DIFF
--- a/uncode/wpml-config.xml
+++ b/uncode/wpml-config.xml
@@ -169,5 +169,12 @@
                 <attribute>subheading</attribute>
             </attributes>
         </shortcode>
+        <shortcode>
+            <tag>vc_column</tag>
+            <attributes>
+                <attribute encoding="vc_link" type="link">link_to</attribute>
+                <attribute type="link">column_link</attribute>
+            </attributes>
+        </shortcode>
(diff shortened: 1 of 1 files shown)


strategio:
@nicolasviallet, for WPBakery page builder, we also find it in [js_composer](https://github.com/OnTheGoSystems/wpml-config/blob/master/js_composer/wpml-config.xml#L392) and [js_composer-dt-the7](https://github.com/OnTheGoSystems/wpml-config/blob/master/js_composer-dt-the7/wpml-config.xml#L392C13-L392C33).

Could please also apply the same changes in these other configs? (unless you think it's very specific to Uncode)

Note: it's an old discrepancy between these 3 files that we have to deal with now.

nicolasviallet:
@strategio we tested it for Visual Composer and it doesn't provide that feature.  It's specific to Uncode theme.

strategio (approved):
Sure @nicolasviallet, then it's fine +1

I'll merge it.

strategio:
Deployed.

PR #415, opened 2024-12-19
[wcml-4918] Copy the field storing the booking buffer period.

decodekult:
Besides the field value, WooCommerce Bookings stores the available slots for bookings in specific transients. Those transienmts get cleared when a booking product is edited, but we still need to clear the transients for the product translations.

This should get merged when WCML 5.4.0 gets released - there, we will include the logic to clear translations transients.


strategio:
@decodekult, I've added a "post release" task in our [release ticket](https://onthegosystems.myjetbrains.com/youtrack/issue/wcml-4661/). I understood we need to merge and deploy that once we rolled out to 100%. Feel free to adjust the "post release" task if it should be done at a different stage.

dgwatkins (approved):
Looks good @decodekult

strategio:
Merged and deployed!

PR #416, opened 2024-12-30
Remove SiteOrigin XML definitions

dgwatkins:
SiteOrigin is maintaining these files in their plugins and they aren't doing anything here.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-430/


strategio:
@dgwatkins, do you know when SiteOrigin started maintaining their XML config locally?

I am asking because we made some changes changes here in the 2 last years => https://github.com/OnTheGoSystems/wpml-config/commits/master/siteorigin/wpml-config.xml

So if we made changes here, I believe it's because it was applied (we should have tested it manually) and as a consequence, the local config was missing.

dgwatkins:
> @dgwatkins, do you know when SiteOrigin started maintaining their XML config locally?

We aren't overriding their local XML, so all these changes had no effect @strategio

dgwatkins:
> I am asking because we made some changes changes here in the 2 last years => https://github.com/OnTheGoSystems/wpml-config/commits/master/siteorigin/wpml-config.xml

Those changes have already been pushed and merged to the author's repo.

strategio:
@dgwatkins, I could see that we were not overriding the XML config.
But you didn't reply to my question: If we made changes to our remote repo less than 1 year ago, it means it was taking effect, right?
So my question again: When did SiteOrigin started providing a local XML config?

dgwatkins:
> @dgwatkins, I could see that we were not overriding the XML config. But you didn't reply to my question: If we made changes to our remote repo less than 1 year ago, it means it was taking effect, right? 

Theoretically yes, but we must have done something wrong because they added their XML configuration more than 3 years ago:

https://github.com/siteorigin/siteorigin-panels/commits/develop/wpml-config.xml
https://github.com/siteorigin/so-widgets-bundle/commits/develop/wpml-config.xml

strategio (approved):
+1 @dgwatkins 

(I left a [comment in the ticket](https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-430/Put-a-bit-of-order-in-SiteOrigin-XML-files#focus=Comments-102-678262.0-0))

PR #419, opened 2025-02-05
Gutenberg - Gallery block lightbox image is not replaced

nicolasviallet:
Adding Media translation support for Gutenberg core/image widget

From: https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlsupp-12412

### DIFF
--- a/sitepress-multilingual-cms/wpml-config.xml
+++ b/sitepress-multilingual-cms/wpml-config.xml
@@ -92,6 +92,7 @@
 			<xpath label="Button URL" type="link">//a/@href</xpath>
 		</gutenberg-block>
 		<gutenberg-block type="core/image" translate="1">
+			<key type="post-ids" name="id" />
 			<xpath label="Caption">//figure/figcaption</xpath>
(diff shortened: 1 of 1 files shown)


sitepress-multilingual-cms/wpml-config.xml
```diff
@@ -92,6 +92,7 @@
 			<xpath label="Button URL" type="link">//a/@href</xpath>
 		</gutenberg-block>
 		<gutenberg-block type="core/image" translate="1">
+			<key type="post-ids" name="id" />
```
dgwatkins, line 95: I think you a subtype attribute too @nicolasviallet 
For example, if its an attachment id, you can use:
```suggestion
			<key type="post-ids" subtype="attachment" name="id" />
```

dgwatkins: @strategio can you confirm the above?

dgwatkins (approved):

dgwatkins:
Merged and deployed :+1:

PR #422, opened 2025-03-06
Yoast SEO - Gutenberg blocks translation - wpmltriage-2557

nicolasviallet:
Yoast introduced Gutenberg blocks. They are already translatable in ATE. However, when opening the translation in the WP editor, the block will show a "Block contains unexpected or invalid content." message. When clicking the "Attempt recovery" button, the block loses its translation. That's because the block is translated but not the block attributes.


yoast-seo/wpml-config.xml
```diff
@@ -41,4 +41,32 @@
...
+        <gutenberg-block type="yoast/faq-block" translate="1">
+            <key name="questions">
+                <key name="*">
+                    <key name="question" />
+                    <key name="answer" />
+                    <key name="jsonQuestion" />
+                    <key name="jsonAnswer" />
+                </key>
+            </key>
+            <xpath>//div/strong[contains(@class,"schema-faq-question")]</xpath>
```
strategio, line 54: @nicolasviallet, for the `xpath` entries, we can use a `label` attribute.

See https://wpml.org/documentation/support/language-configuration-files/make-custom-gutenberg-blocks-translatable/

Could you please add it and check the rendering in ATE? (we should have human readable field labels)

strategio (commented):
@nicolasviallet, besides my comment, I noticed that we have a weird approach for maintaining Yoast SEO configs (see [compdev-701](https://onthegosystems.myjetbrains.com/youtrack/issue/compdev-701/)).

I will keep your PR on hold until I fix it.

strategio:
@nicolasviallet, I've merged the 3 WordPress SEO configurations into 1 (https://github.com/OnTheGoSystems/wpml-config/pull/423).

Please re-build your branch with your changes. Now all configuration changes should go to `wordpress-seo` (free or pro).

PR #424, opened 2025-03-11
[wpmlpb-443] Add support for avada forms conditionals

dgwatkins:
The plan is to include the code to support this in WPML 4.7.3 (not released).
However, we can merge in advance because our current code ignores encoding.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-443


dgwatkins:
@OnTheGoSystems/compatibility needs reviewing together with https://git.onthegosystems.com/glue-plugins/wpml/wpml-page-builders/-/merge_requests/249

strategio:
>However, we can merge in advance because our current code ignores encoding.

@dgwatkins, did you try this config with the stable code?

I understand the "options" codec will not be affected, but then "logics" will now also be handled by the "options" code in the current stable version.

dgwatkins:
We'll have to wait for release.

As soon as the form has a conditional, we get a fatal error:
> PHP Fatal error:  Uncaught Error: Cannot use object of type stdClass as array

dgwatkins:
I still need you approval @strategio

strategio:
@dgwatkins, having a fatal error if you don't have the latest WPML version is a bit scary...

I suggest a different approach (if you confirm that the new code can also work with the old configuration):
- Create a separated ticket for this new config, and we'll wait a few weeks after the release (100% buckets) to deploy the new configuration.

strategio (approved):
@dgwatkins, I am approving this PR, to merge some weeks after the full release of WPML PB ([see my comment in YT](https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-595/Avada-Make-empty-encoding-default-to-fusionoptions#focus=Comments-102-695896.0-0))

PR #423, opened 2025-03-10
[compdev-701] Merge all WordPress SEO configurations into only one

strategio:
- The configurations of `wordpress-seo` and `wordpress-seo-premium` were exactly the same.
- The configuration of `yoast-seo` was missing several entries, but I just took the keys insider `wpseo_internallinks` (which I believe is for backward compatibility). Indeed, this config has not been updated since 2018.

https://onthegosystems.myjetbrains.com/youtrack/issue/compdev-701/

PR #425, opened 2025-03-12
Yoast SEO - Gutenberg blocks translation - wpmltriage-2557

nicolasviallet:
Yoast introduced Gutenberg blocks, which are already translatable in ATE. However, when opening the translation in the WP editor, a "Block contains unexpected or invalid content" message appears. Clicking the "Attempt recovery" button causes the block to lose its translation. This happens because the block is translated, but not the block attributes. The issue is caused by the mismatch between the block's attributes and its content.

Following Pierre's suggestion, I removed classes where possible to simplify the code and broaden its scope.

_Comes from [#422](https://github.com/OnTheGoSystems/wpml-config/pull/422), resubmission after [#423](https://github.com/OnTheGoSystems/wpml-config/pull/423) was merged._

### DIFF
--- a/wordpress-seo/wpml-config.xml
+++ b/wordpress-seo/wpml-config.xml
@@ -64,5 +64,32 @@
 			<key name="breadcrumbs-404crumb" />
 		</key>
 	</admin-texts>
+	<gutenberg-blocks>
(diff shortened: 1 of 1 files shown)


strategio:
Deployed to the CDN.

PR #426, opened 2025-03-13
[wpmlpb-596] Elementor: Translate the template_id for the loop-grid

dgwatkins:
This won't work until wpmlpb-414 is complete.

Meanwhile, I'll keep this as a draft.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-596/Elementor-Convert-the-ID-from-templateid-in-loop-grid-widget

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -384,6 +384,7 @@
                 <field type="Loop Grid: Load more">text</field>
                 <field type="Loop Grid: Load more/No posts custom message">load_more_no_posts_custom_message</field>
                 <field type="Loop Grid: Nothing found message">nothing_found_message_text</field>
+                <field type="post-ids" sub-type="elementor_library">template_id</field>
             </fields>
         </widget>
         <widget name="lottie">
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -386,6 +386,9 @@
                 <field type="Loop Grid: Nothing found message">nothing_found_message_text</field>
                 <field type="post-ids" sub-type="elementor_library">template_id</field>
             </fields>
+            <fields-in-item items_of="alternate_templates">
+                <field type="post-ids" sub-type="elementor_library">template_id</field>
+            </fields-in-item>
         </widget>
         <widget name="lottie">
             <fields>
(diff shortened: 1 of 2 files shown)


wpml-config.xsd
```diff
@@ -25,6 +25,7 @@
             <xs:element type="wpml-custom-term-fields" name="custom-term-fields" minOccurs="0"/>
             <xs:element type="wpml-custom-fields-texts" name="custom-fields-texts" minOccurs="0"/>
             <xs:element type="wpml-admin-texts" name="admin-texts" minOccurs="0"/>
+            <xs:element type="wpml-allow-translatable-job-fields" name="allow-translatable-job-fields" minOccurs="0"/>
```
strategio, line 28: Why do we have this change @dgwatkins?

It was missing in this file? If yes, could you please fix it in another PR?

strategio (commented):
Looks good @dgwatkins, let's use the same attributes as for GB blocks => https://wpml.org/documentation/support/language-configuration-files/make-custom-gutenberg-blocks-translatable/#converting-ids

I am not approving this PR as I understand it's a POC.

elementor/wpml-config.xml
```diff
@@ -384,6 +384,7 @@
                 <field type="Loop Grid: Load more">text</field>
                 <field type="Loop Grid: Load more/No posts custom message">load_more_no_posts_custom_message</field>
                 <field type="Loop Grid: Nothing found message">nothing_found_message_text</field>
+                <field type="post-ids" sub-type="elementor_library">template_id</field>
```
strategio, line 387: @dgwatkins, now I see the `type` attribute is also used for "titles"... It's not optimal, but except using a new attribute that would be even more confusing, I believe we should stick with the `type` attribute.

You will have to desambiguate that in config parser.

dgwatkins: I added a step in wpmlpb-414 to disambiguate.

wpml-config.xsd
dgwatkins: We have a separate commit now, resolving.

dgwatkins:
> I am not approving this PR as I understand it's a POC.

Its not really a POC @strategio 

Its a real case (wpmlpb-596) but it won't work until wpmlpb-414 is complete.

I'll keep it in draft and possibly merge after release, but it can already be approved if you like.

strategio:
@dgwatkins, I prefer that you ping me again when the implementation is complete.

dgwatkins:
> @dgwatkins, I prefer that you ping me again when the implementation is complete.

Of course, I didn't really ping you here. I'm not sure why you reviewed early. But I just followed up with your review.

dgwatkins:
I am handling this merge request in #429 to simplify deployment.

PR #428, opened 2025-03-18
[wpmlpb-588] Elementor: Convert chat button icon for Floating Buttons

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-588


dgwatkins:
We need to merge #426 before this one (for the XSD to get updated)

dgwatkins:
I'm going to move these changes to #429 so we can handle the merge in one go.

It will be a separate commit to keep things organized.

PR #434, opened 2025-03-31
[compdev-4471] Update `wpml-config.xsd` from WPML 4.7.3

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmldev-4471/


strategio:
It's the same changes as in https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/-/merge_requests/5061 (already reviewed) and the CI checks are also passing here.

So I'll merge in `master` right away (no need to deploy yet).

PR #435, opened 2025-04-02
[wpmlpb-570] Beaver Builder: Removing the pricing table widget

diiegopereira:
We are now handling it with a compatibility class (because of the data shape).
See wpmlpb-348.


strategio (approved):
+1 @diiegopereira

PR #436, opened 2025-04-03
[compsupp-7950] Cornerstone Button > Secondary button text not translatable

nicolasviallet:
Improving support for Cornerstone button widget
The Secondary Button Text `anchor_text_secondary_content` is not translatable.

https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-7950

PS: I checked the [cornerstone remote XML](https://github.com/OnTheGoSystems/wpml-config/blob/d1bb06915a0ab9a962405f947bc5bc24ce531a6c/cornerstone/wpml-config.xml#L255) already and it's up-to-date.

### DIFF
--- a/pro/wpml-config.xml
+++ b/pro/wpml-config.xml
@@ -480,6 +480,7 @@
       <fields>
         <field type="Button: Button Text" editor_type="LINE">anchor_text_primary_content</field>
         <field type="Button: Link URL" editor_type="LINK">anchor_href</field>
+        <field type="Button: Secondary Text" editor_type="LINE">anchor_text_secondary_content</field>
       </fields>
     </widget>    
   </cornerstone-widgets>


strategio:
Merged and deployed.

@nicolasviallet, after merging, I noticed you made a change in the **Pro** config, but we also have a **Cornerstone** config with the same `button` widget (which is [already updated](https://github.com/OnTheGoSystems/wpml-config/blob/master/cornerstone/wpml-config.xml#L255)). Perhaps we should merge these 2 config files and have only one source of truth. Could you just check if there are a lot of differences between these 2 files?

strategio:
Thanks for the feedback @nicolasviallet.

Unfortunately we cannot create XML config dependencies right now.
If you think it's worth having one single file including both configs or just make a batch update to the outdated Pro config, feel free to create a new ticket.

It may save us some issues, but I'll leave the decision to you.

dgwatkins:
Doesn't Pro work with the Cornerstone theme?

If so, we would have both XML's loaded.

PR #440, opened 2025-04-07
[compdev-560] Add remote notice for LearnDash on the TM Dashboard

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/compdev-560

### DIFF
--- a/global-wpml-notices/wpml-config.xml
+++ b/global-wpml-notices/wpml-config.xml
@@ -1,2 +1,18 @@
 <wpml-config>
+    <notices>
+        <notice type="info" dismissible="1" id="compdev-560-learndash">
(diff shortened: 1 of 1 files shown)


dgwatkins (approved):
The YT URL is redundant, but if you like to keep its fine.

I would also be against the inline CSS, but I imagine we aren't on time to do anything else.

All in all, it looks good @strategio

strategio:
@dgwatkins 

>The YT URL is redundant, but if you like to keep its fine.

I did that on purpose, we need to have a quick access to the source ticket.

>I would also be against the inline CSS, but I imagine we aren't on time to do anything else.

I've tried to align the paragraph with the top left icon. But it depends on how you design the notice (with a `h2` or `p`).

strategio:
Merged and deployed.

PR #442, opened 2025-04-11
[wpmlwpseo-295] Rank Math SEO: Support term meta translation

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlwpseo-295


dgwatkins (approved):
:+1: @strategio

PR #443, opened 2025-04-14
[comp-3831] Divi 5 - New Blocks XML

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/comp-3831

### DIFF
--- a/Divi/wpml-config.xml
+++ b/Divi/wpml-config.xml
@@ -581,8 +581,698 @@
             </attributes>
         </shortcode>
     </shortcodes>
-    <!-- Fixes an issue with Divi, Extra and Divi Builder. See compsupp-3566 and compsupp-3605 -->
     <gutenberg-blocks>
+        <gutenberg-block type="divi" translate="1">
+            <key name="module">
+                <key name="advanced">
+                    <key name="link">
+                        <key name="*">
+                            <key name="value">
+                                <key type="link" name="url" />
+                            </key>
+                        </key>
+                    </key>
+                </key>
+                <key name="decoration">
+                    <key name="background">
+                        <key name="*">
+                            <key name="value">
+                                <key name="image">
+                                    <key name="url" />
+                                </key>
+                                <key name="video">
+                                    <key name="mp4" />
+                                    <key name="webm" />
+                                </key>
+                            </key>
+                        </key>
+                    </key>
+                </key>
+                <key name="meta">
+                    <key name="adminLabel">
+                        <key name="*">
+                            <key name="value" />
+                        </key>
+                    </key>
+                </key>
+            </key>
+        </gutenberg-block>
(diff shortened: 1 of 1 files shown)


Divi/wpml-config.xml
```diff
@@ -584,5 +584,1063 @@
     <!-- Fixes an issue with Divi, Extra and Divi Builder. See compsupp-3566 and compsupp-3605 -->
     <gutenberg-blocks>
         <gutenberg-block type="divi/placeholder" translate="0" />
+        <gutenberg-block type="divi/row" translate="1">
+            <key name="module">
+                <key name="advanced">
```
strategio, line 589: @dgwatkins, I see you have repeated config for the `module` key in `divi/row` and `divi/column`.

Did you consider "factorizing" it in the [block namespace config](https://wpml.org/documentation/support/language-configuration-files/make-custom-gutenberg-blocks-translatable/#block-namespace)?

strategio (commented):
@dgwatkins, did you use a specific logic to sort the blocks? Otherwise, and since it's a brand new block config, could you sort it by block name alphabetically?

Divi/wpml-config.xml
dgwatkins: Great advice @strategio 

It has simplified the XML quite a bit.

I was not aware of this feature, maybe we should present to the team in a workshop in case someone else missed it.

dgwatkins:
> @dgwatkins, did you use a specific logic to sort the blocks? Otherwise, and since it's a brand new block config, could you sort it by block name alphabetically?

I sorted them alphabetically.

Divi/wpml-config.xml
```diff
@@ -581,8 +581,705 @@
...
+                            </key>
+                        </key>
+                    </key>
+                </key>
+                <key name="decoration">
+                    <key name="background">
+                        <key name="*">
+                            <key name="value">
+                                <key name="image">
+                                    <key name="url" />
```
dgwatkins, line 608: @strategio I understand that image urls should not appear here. They should be handled by "WPML Media Translation" instead. 

But I tested removing this definition and using MT -> it didn't work.

I guess thats why we have them here.

dgwatkins: I created a ticket to handle when time allows: comp-4035

strategio (approved):
Except my remaining comment about HTML ids, it looks good to me @dgwatkins +1

Divi/wpml-config.xml
```diff
@@ -581,8 +581,705 @@
...
+                            </key>
+                        </key>
+                    </key>
+                    <key name="htmlAttributes">
+                        <key name="*">
+                            <key name="value">
+                                <key name="id" />
+                            </key>
+                        </key>
+                    </key>
```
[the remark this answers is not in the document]
dgwatkins: done

PR #454, opened 2025-05-14
wpmlpb-651 - Add encoding to prevent incorrect HTML entity translation

nicolasviallet:
Add the attribute encoding="allow-html-tags" to prevent apostrophes and ampersands from being wrongly translated into HTML entities.


dgwatkins (approved):

dgwatkins:
Merged and deployed

PR #459, opened 2025-06-20
[comp-3831] Update generic config for blocks in Divi namespace

diiegopereira:
We already had a generic configuration, but I updated it to include other common fields.

This fixed the issue related to the blurb block.

Ref: https://onthegosystems.myjetbrains.com/youtrack/issue/comp-3831


strategio (approved):
+1 @diiegopereira

PR #469, opened 2025-07-24
Image Caption is not translatable

nicolasviallet:
Spectra - Ultimate Addons for Gutenberg
Make the Image caption translatable in ATE for the "Image with caption (Spectra)" block.

https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-8089


dgwatkins (approved):
:+1: @nicolasviallet

PR #472, opened 2025-08-01
[wpmlpb-571] Add - Gutenberg - block `page-list` works in other langu…

Pawel-H-OTGS:
…ages

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-571

### DIFF
--- a/sitepress-multilingual-cms/wpml-config.xml
+++ b/sitepress-multilingual-cms/wpml-config.xml
@@ -302,5 +302,8 @@
 				</key>
 			</key>
 		</gutenberg-block>
+		<gutenberg-block type="core/page-list" translate="0">
+			<key name="parentPageID" type="post-ids" sub-type="page" />
+		</gutenberg-block>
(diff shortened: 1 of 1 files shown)


sitepress-multilingual-cms/wpml-config.xml
```diff
@@ -302,5 +302,8 @@
 				</key>
 			</key>
 		</gutenberg-block>
+		<gutenberg-block type="core/page-list" translate="0">
```
Pawel-H-OTGS, line 305: @dgwatkins I have a question, because in the ticket you suggested using:

 `<gutenberg-block type="core/page-list" translate="1">`
 
 but I think "0" is a better fit here - even if I set it to "1", nothing was sent to ATE during the tests

dgwatkins: If you tested, its fine @Pawel-H-OTGS 

It makes sense because there is no content to translate.

PR #473, opened 2025-08-01
[comp-4032] Divi 5-alpha.19.1 - Adding New Blocks XML

andrescifuentesr:
- Adding support for Divi 5-alpha.19.1 new blocks.
- The most important blocks are covered.


andrescifuentesr:
@strategio, I did a global recheck and factorized new and old blocks. As we discussed it, let's push these changes and recheck after that.

PR #494, opened 2025-11-26
[wpmlpb-665] Avada: translatable form notifications

dgwatkins:
We will merge these XML changes with WPML 4.10.

This will give time to existing Avada users to update to 4.9 which will apply this XML configuration via code. 

When we release 4.10, all users should have the updates required for this XML to work and we can remove the code thats applies this configuration and use the XML instead.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-665

### DIFF
--- a/avada/wpml-config.xml
+++ b/avada/wpml-config.xml
@@ -224,6 +224,16 @@
 			<key name="button_1" />
 			<key name="button_2" />
 			<key name="vimeo_id" />
+			<key name="notifications">
+				<key name="*">
+					<key name="label" />
+					<key name="email_subject" />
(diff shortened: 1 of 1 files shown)


strategio:
It looks good @dgwatkins, but I'll give my approval later, so we avoid to mistakenly merge/deploy at the wrong moment.

dgwatkins:
This should be working with 4.9 and safe to deploy now.

Can you approve? I'll merge/deploy and run the CC test.

strategio (approved):
+1 @dgwatkins

PR #518, opened 2026-03-25
[wpmlpb-787] Elementor v4: move e-component to PHP so we can add an integration class

dgwatkins:
It is safe to deploy because we do `class_exists( $class_or_instance )`

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-787/

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -756,24 +756,21 @@
         <!-- WIP: Elementor Editor V4 -->
         <widget name="e-heading">
             <fields>
-                <field type="Heading: Title" field_id="title">title>value</field><!-- V4 beta -->
-                <field type="Heading: Title" field_id="title">title>value>content>value</field>
+                <field type="Heading: Title">title>value>content>value</field>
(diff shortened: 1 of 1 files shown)


strategio (commented):
@dgwatkins, I believe it's fine to remove the V4 beta config now.

But I think it's too early to remove the `e-component` config, WPML 4.9.3 is still in a bucket phase.

Is there a problem if we keep it for longer?

dgwatkins:
> Is there a problem if we keep it for longer?

@strategio the only problem will be for 4.9.3 users trying to translate e-component overridden properties.

dgwatkins:
@strategio we could wait for the announcement, I haven't seen it yet.

Maybe even mention it in the announcement?

strategio:
@dgwatkins, I suggest waiting for WPML 4.9.3 to be released to 100% users (it should not be long). Then deploy that change and close our errata.

dgwatkins:
4.9.4 was released to 100%

@strategio can I merge here now?

strategio (approved):
+1 @dgwatkins

PR #523, opened 2026-04-27
[compsupp-8352] Add configuration for Divi shop block

nicolasviallet:
Added config for the Divi 5 shop block to translate product categories automatically.

Divi Woo Products module was not filtering product categories correctly. 

Ref: https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-8352

### DIFF
--- a/Divi/wpml-config.xml
+++ b/Divi/wpml-config.xml
@@ -1399,6 +1399,19 @@
                 </key>
             </key>
         </gutenberg-block>
+        <gutenberg-block type="divi/shop" translate="0">
+            <key name="content">
+                <key name="advanced">
+                    <key name="includeCategories">
+                        <key name="*">
+                            <key name="value">
+                                <key name="*" type="taxonomy-ids" sub-type="product_cat" />
+                            </key>
+                        </key>
+                    </key>
+                </key>
+            </key>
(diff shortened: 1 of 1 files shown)


Divi/wpml-config.xml
```diff
@@ -1399,6 +1399,19 @@
                 </key>
             </key>
         </gutenberg-block>
+        <gutenberg-block type="divi/shop" translate="0">
```
dgwatkins, line 1402: @nicolasviallet is `translate="0"` correct here? 

Did you test it?

nicolasviallet: This block should remain identical in the translated version, there is no text to translate. That’s why I set `translate="0"`. We only need to translate the product category IDs.
I tested it on this page: https://crimson-sampler.sandbox.otgs.work/fr/exemple-de-page/

The sandbox login link is available in the ticket in case you’d like to run tests.

dgwatkins: I had no idea keys would still be translated when translate is zero.

I have learned something new today :-)

dgwatkins (approved):

dgwatkins:
Merged and deployed

PR #524, opened 2026-05-06
[wpmlpb-828] - Media Translation not supported under a wildcard.

nicolasviallet:
It seems that the newly added types `media-ids` and `media-urls` are not supported under a wildcard. I replaced the wild card with the three possible values `desktop`, `tablet`, `phone`.

The issue was reported for the Image block, but I updated two other blocks with the same configuration: `divi/menu` and `divi/gallery`.

Ref: https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-828/Divi-5-Media-Translation-displays-the-original-image

### DIFF
--- a/Divi/wpml-config.xml
+++ b/Divi/wpml-config.xml
@@ -787,8 +787,26 @@
             </key>
             <key name="image">
                 <key name="innerContent">
-                    <key name="*">
-                        <key name="*">
+                    <key name="desktop">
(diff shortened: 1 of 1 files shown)


dgwatkins (approved):

dgwatkins:
Merged and deployed, thanks @nicolasviallet

PR #525, opened 2026-05-14
[compsupp-8365] Add empty_notice attribute to form fields

nicolasviallet:
While the general 'Empty Input Notice' string is translatable in String Translation, the in-form customizable notice for each field is not translatable in ATE when translating the form.

Adding the empty_notice attribute to all fields allowing it resolves the issue.

Ref: https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-8365


dgwatkins (approved):

dgwatkins:
Merged and deployed, thanks @nicolasviallet

PR #531, opened 2026-07-03
[comp-4504] Setting Custom Layouts to Copy  _wpb_post_custom_layout

andrescifuentesr:
Setting `_wpb_post_custom_layout` custom field as copy for WPBakery 9.0
https://onthegosystems.myjetbrains.com/youtrack/issue/comp-4504

### DIFF
--- a/js_composer/wpml-config.xml
+++ b/js_composer/wpml-config.xml
@@ -7,6 +7,7 @@
     <custom-fields>
         <custom-field action="ignore">_vc_post_settings</custom-field>
         <custom-field action="copy">_wpb_shortcodes_custom_css</custom-field>
+        <custom-field action="copy">_wpb_post_custom_layout</custom-field>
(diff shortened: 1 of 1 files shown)


strategio:
@andrescifuentesr, I think this change is OK.

It may be blocking if a user wants to have a custom layout while translating directly with the WPBakery editor.
But for now it's better that way (which satisfies the most common workflow).

If we get reports in the future, we could do the following:
- Change this remote config to `copy-once`
- Add logic in WPML PB to copy this field when a translation job is applied.

One last question: Is this field added in WPBakery 9.0 only or did it already exist before?

andrescifuentesr:
@strategio,  `_wpb_post_custom_layout` custom field was already present in WPBakery Page Builder 8.7.3. Then, I created a new sandbox and the issue was already there, so it is not related to the beta. Surprisingly, we haven't gotten any reports.

strategio (approved):

strategio:
Merged and deployed.

PR #534, opened 2026-07-06
[wpmlpb-857] divi/menu: capture rendered logo title/alt from decoration

dgwatkins:
Divi renders the menu logo title/alt from the module decoration attributes (targetElement="logo"), but the config only captured the non-rendered logo>innerContent alt/titleText copies - so translating them had no front-end effect. Add the module>decoration>attributes>* capture (as divi/image already has) and drop the dead logo>innerContent alt/titleText.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-857/

### DIFF
--- a/Divi/wpml-config.xml
+++ b/Divi/wpml-config.xml
@@ -847,14 +847,27 @@
             </key>
         </gutenberg-block>
         <gutenberg-block type="divi/menu" translate="1">
+            <key name="module">
+                <key name="decoration">
+                    <key name="attributes">
+                        <key name="*">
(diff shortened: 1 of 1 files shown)


dgwatkins:
The attributes array is indexed based, I cannot target alt/title attributes directly, but its ok because:
- There are no other attributes in divi/menu
- Although I can't add it to the global namespace, because it would pull in attributes from other divi modules that aren't translatable.

strategio (approved):
+1 @dgwatkins

PR #537, opened 2026-07-07
[wpmlpb-841] Elementor: translate atomic image (e-image) alt text

dgwatkins:
Register the atomic image widget's stored alt (image>value>src>value>alt>value) as a translatable string so it is picked up for translation. This covers URL-based atomic images, whose rendered alt comes from the widget.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-841/


strategio (approved):
+1 @dgwatkins

PR #538, opened 2026-07-13
[wpmlpb-834] Add the Elementor atomic Select form widget

d-packs:
Registers the option labels of the atomic Select widget (e-form-select) for translation. Its options are a non-associative list at options>value, each a key-value entry whose key holds the visible label; requires the WPML Page Builders release with fields-in-item support for nested items fields.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-834

### DIFF
--- a/elementor/wpml-config.xml
+++ b/elementor/wpml-config.xml
@@ -806,6 +806,11 @@
         </widget>
         <widget name="e-form-file-upload">
         </widget>
+        <widget name="e-form-select">
+            <fields-in-item items_of="options>value">
+                <field type="Form Select: Option Label" editor_type="LINE">value>key>value</field>
+            </fields-in-item>
+        </widget>
         <widget name="e-form-submit-button">
             <fields>
                 <field type="Form: Submit Button">label>value</field>


elementor/wpml-config.xml
```diff
@@ -806,6 +806,11 @@
         </widget>
         <widget name="e-form-file-upload">
         </widget>
+        <widget name="e-form-select">
+            <fields-in-item items_of="options>value">
+                <field type="Form: Select option label" editor_type="LINE">value>key>value</field>
```
dgwatkins: I noticed that the fields appear in ATE grouped by widget. There is no grouping done for a form with nested widgets (see [wpmlpb-762]).

For this reason, I think it may be clearer to set this to:
`                <field type="Form Select: Option Label" editor_type="LINE">value>key>value</field>`

And I'd like to mention if we should translate the key too. Maybe bring it up in our daily call today @d-packs

d-packs:
@dgwatkins Makes sense. I made the correction.

dgwatkins (approved):

dgwatkins:
Merged and deployed @d-packs

PR #543, opened 2026-07-21
Register woocommerce-warranty in config-index.xml

SNO7E-G:
Summary: Registers the existing woocommerce-warranty/wpml-config.xml in the index so WPML actually serves it.

  Problem: The config has shipped since 2017 (commit cdb66d7) but was never added to config-index.xml — so it has never been served.

  Change: Two <item id="woocommerce-warranty"> alias lines (WooCommerce Warranty Requests + Warranty Requests), placed alphabetically after
  woocommerce-video-tab, with override_local="true" to match every other woocommerce-* entry.

  Risk: None — registration only; no translate/copy values authored.

  Verified: schema passes · build-index resolves the entry · orphan count 1 → 0.

### DIFF
--- a/config-index.xml
+++ b/config-index.xml
@@ -103,6 +103,8 @@
 		<item id="woocommerce-social-media-share-buttons" override_local="true">Woocommerce Social Media Share Buttons</item>
 		<item id="woocommerce-tab-manager" override_local="true">WooCommerce Tab Manager</item>
 		<item id="woocommerce-video-tab" override_local="true">WooCommerce Video Tab</item>
+		<item id="woocommerce-warranty" override_local="true">WooCommerce Warranty Requests</item>
+		<item id="woocommerce-warranty" override_local="true">Warranty Requests</item>
 		<item id="wordpress-seo" override_local="true">Yoast SEO</item>
 		<item id="wordpress-seo" override_local="true">Yoast SEO Premium</item>
 		<item id="wp-job-manager">WP Job Manager</item>


strategio (approved):
Hi @SNO7E-G, thanks for your contribution.

Unfortunately, I could not find a reference from the commit cdb66d7eb4aa6ff7f3a9e87a8479dc8ab5c0f712, and we don't really know how this was tested. And it's been almost 10 years since the initial config was set, so things might have changed.

However, by the name of the custom fields, I guess this configuration is legit.
And this is a low popularity WC addon with limited impact.
I believe there's more benefits than risks, so I will deploy your changes.

Note that it may take up to 24h until the changes are fully propagated.

Could you please re-test in 48h and let us know if it works as expected?

(merged and deployed)

SNO7E-G:
Hi @strategio , thank you for reviewing and deploying this — I appreciate the quick turnaround.

A few notes on provenance and testing that may help:

- This PR only registers an existing file. woocommerce-warranty/wpml-config.xml has been in the repo since 2017 (commit cdb66d7) but was never referenced in config-index.xml, so it was effectively dormant — never served to sites. The PR adds the missing index entry; the config content is unchanged.
- The five meta keys correspond to the WooCommerce Warranty Requests extension, which is still actively maintained (current v2.7.9) and stores its warranty configuration as product post-meta. The classifications match that model: _warranty and _warranty_label hold user-facing text (→ translate), while _warranty_duration, _warranty_type, and _warranty_unit hold structural values (→ copy).
- I validated locally against the current schema (composer run-script validate-xml) and build-index; both pass.

I'll verify after the 48h propagation window and report back. If it helps, I'm glad to run a specific scenario on a test site — just let me know what you'd like checked. Thanks again

PR #546, opened 2026-07-22
Divi / Divi Builder: remove duplicate _et_pb_show_page_creation custom-field

SNO7E-G:
**Summary:** Removes a duplicated custom-field that appears in both the Divi theme and the Divi Builder plugin configs.

**Problem:** `_et_pb_show_page_creation` (`action="copy"`) is listed twice in the `<custom-fields>` block of both `Divi/wpml-config.xml` and `divi-builder/wpml-config.xml` — a copy-paste carried across the two sibling configs.

**Change:** Deletes the redundant second occurrence in each file (one commit, two files).

**Risk:** None — the two entries are byte-identical, so the field stays registered once with the same action; applying an identical directive once instead of twice yields the same result.

**Verified:** schema passes · build-index passes · each file lists the key once · diff is two deletions.


dgwatkins (approved):
Thanks @SNO7E-G

PR #548, opened 2026-07-23
Avada: remove duplicate entries

SNO7E-G:
**Summary:** Consolidates several duplicated entries across the Avada config into single occurrences.

**Problem:** Multiple entries are listed twice: custom-field `pyre_project_url_text` (`action="translate"`); the `fusion_dropcap` `<shortcode>` definition; four admin-text keys under `fusion_options` (`email_address`, `blog_subtitle`, `date_format`, `sharing_social_tagline`); and two `<shortcode-list>` items (`fusion_popover`, `fusion_soundcloud`).

**Change:** Removes the redundant second occurrence of each. The `fusion_dropcap` reference inside `<shortcode-list>` is a separate, valid registration and is kept.

**Risk:** None — every removal keeps one identical entry.

**Verified:** schema passes · build-index passes · each item appears once · diff is removals only.


dgwatkins (approved):
Thanks @SNO7E-G 

You're on a rollercoaster! 😄

SNO7E-G:
😅 Just trying to keep the momentum going! 
Thanks for the quick reviews and merges.

PR #549, opened 2026-07-23
Brooklyn: remove duplicate shortcode definitions

SNO7E-G:
**Summary:** Removes two fully duplicated `<shortcode>` blocks from the Brooklyn theme config.

**Problem:** `ut_service_column` and `ut_service_column_vertical` are each defined twice with an identical tag and `<attribute>headline</attribute>`.

**Change:** Deletes the redundant second `<shortcode>` block for each (12 lines total).

**Risk:** None — identical definitions; each shortcode stays registered once with the same attributes.

**Verified:** schema passes · build-index passes · each shortcode defined once · diff is 12 deletions.


dgwatkins (approved):
Thanks @SNO7E-G

PR #550, opened 2026-07-23
WooCommerce Composite Products: remove duplicate _bto_selection_mode custom-field

SNO7E-G:
**Summary:** Removes a duplicated custom-field from the WooCommerce Composite Products config.

**Problem:** `_bto_selection_mode` (`action="copy"`) is listed twice in the `<custom-fields>` block (lines 4 and 8).

**Change:** Deletes the redundant second occurrence (line 8), keeping the first (line 4).

**Risk:** None — identical entries.


dgwatkins (approved):
Thanks @SNO7E-G

PR #551, opened 2026-07-23
X theme: remove duplicate shortcode-list entries

SNO7E-G:
**Summary:** Removes duplicate tokens from the X theme `<shortcode-list>`, which was almost entirely doubled.

**Problem:** The `<shortcode-list>` contains 42 duplicated tokens — the list is nearly a full repetition of itself (89 tokens, only 47 unique).

**Change:** Removes each redundant repeat, keeping one occurrence of each in original order (89 → 47 tokens).

**Risk:** None — an inert token list; duplicate tokens have no behavioural effect.

**Verified:** schema passes · build-index passes · no duplicate tokens remain · diff is the `<shortcode-list>` line only.


dgwatkins (approved):
Thanks @SNO7E-G

PR #552, opened 2026-07-23
MailChimp for WordPress: remove duplicate text_subscribed custom-field

SNO7E-G:
**Summary:** Removes a duplicated custom-field from the MailChimp for WordPress config.

**Problem:** `text_subscribed` (`action="translate"`) is listed twice in the `<custom-fields>` block (lines 6 and 13).

**Change:** Deletes the redundant second occurrence (line 13), keeping the first (line 6).

**Risk:** None — identical entries.


strategio (approved):
+1 @SNO7E-G

strategio:
Merged and deployed.

PR #558, opened 2026-07-28
Enfold: remove duplicate shortcode-list entries

SNO7E-G:
**Summary:** Removes a duplicate token from the Enfold `<shortcode-list>`.

**Problem:** The `<shortcode-list>` includes `av_testimonial_single` twice.

**Change:** Removes the redundant repeat, keeping one occurrence in original order. (Independent of — and does not touch — the unresolved `av_testimonial_single` `<tag>` conflict noted under "Flagged".)

**Risk:** None — an inert token list; a duplicate token has no behavioural effect.

**Verified:** schema passes · build-index passes · no duplicate tokens remain · diff is the `<shortcode-list>` line only.


dgwatkins (approved):
👍 @SNO7E-G

PR #559, opened 2026-07-30
EDD: fix conflicting translation action for edd_variable_prices

SNO7E-G:
**Summary:** Resolves a contradictory declaration so Easy Digital Downloads variable-price option names are translated correctly.

**Problem:** `edd_variable_prices` is declared twice with conflicting actions — `action="copy"` **and** `action="translate"` — alongside a `<custom-fields-texts>` block that scopes translation to the `name` sub-key. WPML can only apply one action, so the intent is ambiguous.

**Change:** Removes the stray `action="copy"`, leaving a single `action="translate"` plus the existing `<custom-fields-texts>` block — exactly matching the identically-structured sibling `edd_download_files` in the same file.

**Risk:** Behaviour change (intended) — price-option *names* become translatable; amounts and all other sub-keys are preserved (not registered for translation). Backed by a repo-wide check: every `<custom-fields-texts>` key pairs with `action="translate"` (avada `_fusion`, CF7 `_mail`/`_mail_2`, `edd_download_files`, sitepress `footnotes`, Yoast `_yoast_wpseo_focuskeywords`) — `edd_variable_prices` was the only outlier. Schema validation cannot confirm this one (both `copy` and `translate` validate), so correctness rests on the convention evidence. Happy for maintainers to make the final call or convert to an issue.

**Verified:** schema passes · build-index passes · the key now resolves to a single `translate` + the texts-block · diff is one deletion.


dgwatkins (approved):
:+1: @SNO7E-G

PR #566, opened 2026-09-08
Convert taxonomy include/exclude IDs on Post Card element - compsupp-8447

nicolasviallet:
Match include_*/exclude_* fields via regex instead of hardcoding each taxonomy, so any taxonomy (category, product_cat, portfolio_category, faq_category, and custom ones) gets its IDs converted automatically.

Fields that aren't taxonomy ID lists (product_skus, base64-encoded data) are left untouched.

Ref: [https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-8447](https://onthegosystems.myjetbrains.com/youtrack/issue/compsupp-8447)


dgwatkins (approved):

dgwatkins:
tx @nicolasviallet

PR #567, opened 2026-09-11
[comp-4565] Read Elementor 4.3 atomic text from the escaped-html prop

strategio:
## What this changes

Elementor 4.3 changes the text prop of every atomic element from `html-v3` to `escaped-html`. The saved value moves:

```
4.2:  "title": {"$$type":"html-v3","value":{"content":{"$$type":"string","value":"Text"}}}
4.3:  "title": {"$$type":"escaped-html","value":"Text"}
```

The five current paths end in `>value>content>value`, so on 4.3 they resolve to nothing and WPML offers no atomic text for translation — heading, paragraph, button, form label and form submit button.

This adds the `…>value` path next to each current path.

## Why the old paths stay

One config serves every site and cannot branch on the plugin version. Sites on Elementor 4.2, and 4.3 sites whose pages have not migrated yet, still need them. `e-form-submit-button` already carries two paths for one field after the 4.1 `label` → `text` rename (#540), so this follows the existing pattern.

## Why `field_id`

`WPML_Elementor_Translatable_Nodes::get_string_name()` builds the string name from `field_id` when set, and falls back to `field`. `Parser.php` already reads a `field_id` attribute from this XML.

Without it the string name would follow the new path, and every already-translated Elementor page would lose its translations the first time Elementor migrates it. Pointing `field_id` at the old path keeps the name identical for both shapes.

## Testing

Elementor Free and Pro 4.3.0-beta2, WPML core develop, this config loaded as the downloaded Elementor config:

| Case | Result |
|---|---|
| Page built in the 4.3 editor | strings registered, names identical to the 4.2 form |
| Page built through Elementor MCP (new Accordion + List) | strings registered |
| All five changed fields, 4.3 shape | all five registered |
| Page in the 4.2 `html-v3` shape | unchanged, no regression |
| Translate under 4.2, migrate to 4.3, re-register | German front end still shows the translation |

No duplicate or empty strings, `debug.log` clean. Elementor ships the `html-v3-to-escaped-html` migration and the change is already in 4.3.0-beta1, so it looks final for the 22 September release.

Ticket: https://onthegosystems.myjetbrains.com/youtrack/issue/comp-4565/


dgwatkins (approved):
Looks good @strategio

strategio:
Merged and deployed.

